PR #9, opened 2024-02-15
Resnet variants

DrejcPesjak:
Added support for ResNet 18, 34, 50, 101, 152 backbones.
Using _torchvision_ library from **pytorch** ([source](https://pytorch.org/vision/main/models/resnet.html)).

### DIFF
--- a/configs/resnet_model.yaml
+++ b/configs/resnet_model.yaml
@@ -1,0 +1,121 @@
+
+model:
+  name: resnet18_classification
+  nodes:
+    - name: ResNet18
+      download_weights: True
+
+    - name: ClassificationHead
+      inputs:
+        - ResNet18
+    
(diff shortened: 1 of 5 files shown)


klemen1999 (approved):
LGTM but perhaps set a limit to variants based on beanchmarking? Meaning check if some variants are perhaps too big (e.g. 152) to be actually usable on e.g OAK devices.

tersekmatija:
> LGTM but perhaps set a limit to variants based on beanchmarking? Meaning check if some variants are perhaps too big (e.g. 152) to be actually usable on e.g OAK devices.

I think this should be left to the user/EasyML/predefined configs to handle? But yeah unlikely something > RN101 would be used.

kozlov721 (approved):
LGTM

PR #10, opened 2024-02-16
MLFlow Upload Fix

klemen1999:
- Fixed `ExportOnTrainEnd` callback's ability to upload to mlflow (also some fixes in `Exporter` related to this)
- Edited `UploadCheckpoint` callback (previously `UploadCheckpointOnTrainEnd`) so it now uploads currently best checkpoint (based on validation loss) after every validation interval. This way we have some intermediate checkpoint in case training crashes before it is finished.


jkbmrz (approved):
Works as expected on my end!

kozlov721 (approved):
LGTM

PR #11, opened 2024-02-20
CLI Source Option

kozlov721:
# CLI Source Option

Added option to source custom python file before any command is run with CLI.

This makes it possible to use custom components together with the CLI. 

## Usage

**custom.py:**
```python
from luxonis_train.attached_modules.losses import BaseLoss

class CustomLoss(BaseLoss):

    # imports must be inside the class for the sourcing to work properly
    from torch import Tensor

    from luxonis_train.utils.types import Labels, Packet

    def __init__(self, **kwargs):
        super().__init__(**kwargs, required_labels=[])

    def prepare(self, inputs: Packet[Tensor], _: Labels) -> tuple[Tensor]:
        return (inputs["features"][-1],)

    def forward(self, x: Tensor) -> Tensor:
        return x.mean()
```

**config.yaml:**
```yaml
model:
  name: resnet18_classification
  nodes:
    - name: ResNet
      variant: "18"
      download_weights: True

  losses:
    - name: CustomLoss
      attached_to: ResNet

```

**CLI:**
```
luxonis_train --source custom.py train --config config.yaml
```


conorsim (approved):
LGTM

PR #12, opened 2024-02-21
Fix Removed Tensor Metadata

kozlov721:
Fixed a bug when the tensor metadata (like `requires_grad`) are being removed when incorrectly using match-case statement.

**Incorrect way:**
```python
match x:
    case Tensor(data=tensor):
        return tensor
```

This only returns the underlying data in a clean new `Tensor`.

**Correct way:**
```python
match x:
    case Tensor() as tensor:
        return tensor
```

This returns the actual original `Tensor`.


conorsim (commented):
LGTM

PR #14, opened 2024-02-22
Automatic Inference of attach_index

kozlov721:
# Automatic `attach_index`

Added an option to automatically infer sensible attach index based on the type signature of `forward` method. The `attach_index` therefore don't have to be defined by user if correct type annotations are present. 

If the `forward` method has an annotated `inputs`, `input`, or `x` parameter, the type annotation is used to infer the correct `attach_index`. This will be either `"all"` for `inputs: list[Tensor]` or `-1` for `inputs: Tensor`. `"all"` also stays the default if unable to infer or not specified.

### DIFF
--- a/luxonis_train/nodes/base_node.py
+++ b/luxonis_train/nodes/base_node.py
@@ -96,8 +96,9 @@ class BaseNode(
         super().__init__()
 
         if attach_index is None:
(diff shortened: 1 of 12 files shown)


conorsim (approved):
LGTM

PR #17, opened 2024-03-13
Generate NN archive from training configs

jkbmrz:
This PR introduces a new generator class (`luxonis_train.core.Archiver`) enabling automatic generation of NN archives from training configs. This functionality is exposed both through a command-line interface (CLI) and callbacks (with uploading the NN archive to MLFlow). Furthermore, basic test coverage is included to ensure correctness of generation.

The added features are functional but there are still a few open issues so any insights or suggestions regarding the current implementation would be greatly appreciated!

Open issues:
- (**model formats**): Currently, the generator is only able to archive for ONNX model format. Do we want to support it for other formats also (e.g. BLOB, DLC, XML/BIN… )? If so, how to approach models with multiple executables (e.g. the [path parameter](https://github.com/luxonis/luxonis-ml/blob/dev/luxonis_ml/nn_archive/config_building_blocks/base_models/metadata.py) in NN archive `Metadata` class requires a single path to the model executable so I'm not sure what to do in case of multiple)?
- (**is_softmax** parameter, e.g. in `HeadClassification` [class](https://github.com/luxonis/luxonis-ml/blob/10f9a4323c3ac52e946ef4ba5f49f1701ea0b2a7/luxonis_ml/nn_archive/config_building_blocks/base_models/head.py#L80)): Currently, decision if output is already softmaxed is based on iterating ONNX nodes and checking if there is a node named softmax. However, I'm not sure how bulletproof is that and if it is implementable for other model formats. Is there some other way to determine this parameter (OR should maybe even be hard-coded for specific architectures)?
- (**head_outputs** [classes](https://github.com/luxonis/luxonis-ml/blob/dev/luxonis_ml/nn_archive/config_building_blocks/base_models/head_outputs.py)): I’m wondering how to determine which outputs belong to specific head output parameters (cc @conorsim). For example:
  - `yolo_outputs` parameter in `EfficientBBoxHead`. Is it sensible to expect that all outputs of the model always belong here? If not, how to determine the outputs to be listed?
  - `predictions` parameter in `ClassificationHead`. Is it sensible to expect that models with `ClassificationHead` will always only have one output? If not, how to determine which of the outputs is the right one?
  - `boxes` and `scores` parameters in `ObjectDetectionSSD`. How to determine which output belongs to which parameter for an arbitrary SSD network?
- (**tests**): I’ve implemented setup class method that makes a dummy LDF, config, and ONNX model, and runs the archivation. Next, the tests are run to inspect basic characteristics of the constructed archive file. In the end, teardown class method deletes all the constructed files. Is this approach OK or is there some better way to tackle this?

### DIFF
--- a/luxonis_train/core/archiver.py
+++ b/luxonis_train/core/archiver.py
@@ -1,17 +1,16 @@
 import os
 from logging import getLogger
 from pathlib import Path
-from typing import Any, List, Dict
+from typing import Any
+
 import onnx
-
-from luxonis_train.utils.config import Config
-from luxonis_train.models import LuxonisModel
-from luxonis_train.nodes.enums.head_categorization import ImplementedHeads
-
-from luxonis_ml.utils import LuxonisFileSystem
-
 from luxonis_ml.nn_archive.archive_generator import ArchiveGenerator
 from luxonis_ml.nn_archive.config import CONFIG_VERSION
+from luxonis_ml.utils import LuxonisFileSystem
+
+from luxonis_train.models import LuxonisModel
+from luxonis_train.nodes.enums.head_categorization import ImplementedHeads
+from luxonis_train.utils.config import Config
 
 from .core import Core
 
@@ -19,7 +18,8 @@ logger = getLogger(__name__)
 
 
 class Archiver(Core):
-    """Main API which is used to construct the NN archive file out of a trainig config and model executables."""
+    """Main API which is used to construct the NN archive out of a trainig config and
+    model executables."""
 
     def __init__(
         self,
@@ -30,13 +30,9 @@ class Archiver(Core):
 
         @type cfg: str | dict[str, Any] | Config
         @param cfg: Path to config file or config dict used to setup training.
-
         @type opts: list[str] | tuple[str, ...] | dict[str, Any] | None
         @param opts: Argument dict provided through command line,
             used for config overriding.
-
-        @type executable: str
-        @param executable: Path to model executable file.
         """
 
         super().__init__(cfg, opts)
@@ -49,7 +45,7 @@ class Archiver(Core):
         )
 
         self.model_name = self.cfg.model.name
-        
+
         self.archive_name = self.cfg.archiver.archive_name
         archive_save_directory = Path(self.cfg.archiver.archive_save_directory)
         if not archive_save_directory.exists():
@@ -92,15 +88,9 @@ class Archiver(Core):
         for output_name in outputs_dict:
             self._add_output(name=output_name, dtype=outputs_dict[output_name]["dtype"])
 
-        heads_dict = self._get_heads()
+        heads_dict = self._get_heads(executable_path)
         for head_name in heads_dict:
-            outputs_list = [
-                output_name for output_name in outputs_dict if True
-            ]  # TODO: currently we list all output names. Is this correct?
-            self._add_head(
-                outputs_list=outputs_list,
-                head_metadata=heads_dict[head_name]["head_metadata"],
-            )
+            self._add_head(heads_dict[head_name])
 
         model = {
             "metadata": {
@@ -124,14 +114,10 @@ class Archiver(Core):
             executables_paths=[executable_path],  # TODO: what if more executables?
         ).make_archive()
 
-        #self.archive_path = os.path.join(
-        #    self.archive_save_directory, f"{self.archive_name}.tar.gz"
-        #)  # TODO: instead of making it manually, modify ArchiveGenerator.make_archive() to return it!
-
         if self.cfg.archiver.upload_url is not None:
             self._upload()
 
-        return self.archive_path  # TODO: is this necessary?
+        return self.archive_path
 
     def _get_inputs(self, executable_path: str):
         """Get inputs of a model executable.
@@ -182,27 +168,23 @@ class Archiver(Core):
 
         @type name: str
         @param name: Name of the input layer.
-
         @type dtype: str
         @param dtype: Data type of the input data (e.g., 'float32').
-
         @type shape: list
         @param shape: Shape of the input data as a list of integers (e.g. [H,W], [H,W,C], [BS,H,W,C], ...).
-
         @type preprocessing: dict
         @param preprocessing: Preprocessing steps applied to the input data.
-
         @type input_type: str
         @param input_type: Type of input data (e.g., 'image').
         """
 
         self.inputs.append(
             {
-                "name": name,  # name of the input layer
+                "name": name,
                 "dtype": dtype,
                 "input_type": input_type,
-                "shape": shape,  # Shape of the input data as a list of integers (e.g. [H,W], [H,W,C], [BS,H,W,C], ...).
-                "preprocessing": preprocessing,  # Preprocessing steps applied to the input data.
+                "shape": shape,
+                "preprocessing": preprocessing,
             }
         )
 
@@ -238,11 +220,10 @@ class Archiver(Core):
         return outputs_dict
 
     def _add_output(self, name: str, dtype: str) -> None:
-        """Add output to self.outputs
+        """Add output to self.outputs.
 
         @type name: str
         @param name: Name of the output layer.
-
         @type dtype: str
         @param dtype: Data type of the output data (e.g., 'float32').
         """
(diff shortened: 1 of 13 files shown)


jkbmrz:
Any ideas why are we getting all these import errors during the checks? We could manually add them all to the `requirements.txt` but its strange because imports fail even for some packages not introduced by my PR (e.g. cv2)

luxonis_train/core/archiver.py
```diff
@@ -0,0 +1,396 @@
...
+            head_node = self.lightning_module._modules["nodes"][head_alias]
+            parameters["iou_threshold"] = head_node.iou_thres
+            parameters["conf_threshold"] = head_node.conf_thres
+            parameters["max_det"] = head_node.max_det
+            # head_outputs["n_keypoints"] # TODO: implement
+            # head_outputs["n_prototypes"] # TODO: implement
+            # head_outputs["prototype_output_name"] # TODO: implement
+        elif head_name == "ObjectDetectionSSD":
+            raise NotImplementedError
+            # head_outputs["anchors"] # TODO: implement
```
conorsim, line 274: I don't think we support any architectures that will require anchors at the moment, just FYI

luxonis_train/core/archiver.py
```diff
@@ -0,0 +1,396 @@
...
+        elif head_name == "SegmentationHead":
+            parameters["is_softmax"] = self._is_softmax(executable_path)  # TODO: test
+        elif head_name == "BiSeNetHead":
+            parameters["is_softmax"] = self._is_softmax(executable_path)  # TODO: test
+        elif head_name == "ImplicitKeypointBBoxHead":
+            raise NotImplementedError
```
conorsim, line 280: I think we should be able to implement this one now with latest updates. The ONNX for `ImplicitKeypointBBoxHead` should just have one output, which will be `predictions` in NN archive.

luxonis_train/nodes/efficient_bbox_head.py
```diff
@@ -30,6 +30,7 @@ def __init__(
         n_heads: Literal[2, 3, 4] = 3,
         conf_thres: float = 0.25,
         iou_thres: float = 0.45,
+        max_det: int = 300,
```
conorsim, line 33: So I think honestly this is a unique parameter of NN archive which is more for depthai or user preference. I don't think it needs to be a head parameter. We could set it as a constant somewhere else and assume users can change it if desired.

CC @kozlov721 for your opinion as well.

conorsim (commented):
Nice, thank you! Just adding some comments for now since there's some questions to sort out. Will reply to those separately.

conorsim:
> (model formats): Currently, the generator is only able to archive for ONNX model format. Do we want to support it for other formats also (e.g. BLOB, DLC, XML/BIN… )? If so, how to approach models with multiple executables (e.g. the [path parameter](https://github.com/luxonis/luxonis-ml/blob/dev/luxonis_ml/nn_archive/config_building_blocks/base_models/metadata.py) in NN archive Metadata class requires a single path to the model executable so I'm not sure what to do in case of multiple)?

In my opinion we only need to support ONNX, especially for now. We might still need to address the secondary executable path issue for ONNX as well, but since we don't have the instance segmentation model implemented at the moment, I wouldn't worry about it in this PR.

conorsim:
> (head_outputs [classes](https://github.com/luxonis/luxonis-ml/blob/dev/luxonis_ml/nn_archive/config_building_blocks/base_models/head_outputs.py)): I’m wondering how to determine which outputs belong to specific head output parameters (cc @conorsim). For example:
yolo_outputs parameter in EfficientBBoxHead. Is it sensible to expect that all outputs of the model always belong here? If not, how to determine the outputs to be listed?
predictions parameter in ClassificationHead. Is it sensible to expect that models with ClassificationHead will always only have one output? If not, how to determine which of the outputs is the right one?
boxes and scores parameters in ObjectDetectionSSD. How to determine which output belongs to which parameter for an arbitrary SSD network?

yolo_outputs: Yes, I think is it sensible to expect that all outputs of the model always belong here.

predictions: Yeah this is the "default" configuration in NN archive where there is only one ONNX output. I think it's safe to assume only one ONNX output in this case.

boxes and scores: Let's not worry about this for now since it's not implemented in luxonis-train.

conorsim:
> (tests): I’ve implemented setup class method that makes a dummy LDF, config, and ONNX model, and run the archivation within the. Next, the tests are run to inspect basic characteristics of the constructed archive file. In the end, teardown class method deletes all the constructed files. Is this approach OK or is there some better way to tackle this?

Yeah this seems good to me. It will be nice to also add tests for each type of head in luxonis-train once other questions are resolved.

luxonis_train/core/archiver.py
kozlov721: `ImplicitKeypointBBoxHead` head does require anchors.

luxonis_train/core/archiver.py
```diff
@@ -0,0 +1,396 @@
...
+    def _is_softmax_onnx(self, executable_path) -> bool:
+        """Check if ONNX model output is softmaxed.
+
+        @type executable_path: str
+        @param executable_path: Path to model executable file.
+        """
+
+        model = onnx.load(executable_path)
+        for node in model.graph.node:
+            if node.op_type.lower() == "softmax":
```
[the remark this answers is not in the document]
kozlov721: We could run the ONNX model and check that the outputs sum to 1.  It's also not an optimal solution though.

luxonis_train/nodes/efficient_bbox_head.py
kozlov721: We can then say the same about `conf_thres` and `iou_thres` as those are all parameters of `nms`. The reason the `nms` parameters are head parameters is to ensure the same parameters will be used for all metrics / visualizers etc. and we don't need to specify it for each of them separately.

kozlov721:
> > Any ideas why are we getting all these import errors during the checks? Adding the missing packages to the `requirements.txt` might help but its strange because imports fail even for some package imports that existed before my PR (e.g. cv2).
> 
> I think this is because you `import cv2` in the tests you added. And I think OpenCV is not actually a requirement of this repo. But CC @kozlov721 as well.

`opencv-python` should be installed as part of `luxonis-ml`, so I'm not sure what's the issue exactly. We can add it also to requirements of `luxonis-train`, but it's strange.

luxonis_train/core/archiver.py
conorsim: Ah thanks for pointing that out. Actually need to make a change to NN archive then

conorsim: (But realizing I need all the object detection parameters in the keypoint head anyways in NN archive)

luxonis_train/core/archiver.py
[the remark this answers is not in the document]
conorsim: Yeah that's not a bad idea either. I think we could hardcode into each classification/segmentation head to keep it simple for now though

luxonis_train/nodes/efficient_bbox_head.py
conorsim: Oh my bad I didn't realize they're used in visualizers - makes sense. Disregard my comment

luxonis_train/core/archiver.py
jkbmrz: adding support in commit [3c0ddc3](https://github.com/luxonis/luxonis-train/pull/17/commits/3c0ddc3088aa834f4b070fdc0808805418dfcc91)

tests/unittests/test_core/test_archiver.py
```diff
@@ -0,0 +1,153 @@
...
+                "archive_save_directory": cls.tmp_path,
+            },
+        }
+        cls.config_path = os.path.join(cls.tmp_path, "tmp_config.yaml")
+        with open(cls.config_path, "w") as yaml_file:
+            yaml_str = yaml.dump(config_dict)
+            yaml_file.write(yaml_str)
+
+        # make model
+        model = torchvision.models.mobilenet_v2(pretrained=False)
```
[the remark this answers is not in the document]
jkbmrz: I agree, will look into it!

kozlov721:
> I've added `opencv-python` to `requirements.txt` but now other packages are reported missing. I'm not sure why but maybe `luxonis-ml` isn't installed properly. Might it be possible that I disrupted something by listing its dev version in `requirements.txt` [here](https://github.com/luxonis/luxonis-train/blob/0e3b4de0d60ca8e185e2de1e4e7b76c5a5dfa91c/requirements.txt#L4)?

You're right, seems you accidentally removed the `[all]` specifier so it only installs `luxonis-ml[utils]`. Should be fixed. 
There's also `mlflow` package missing; hotfix for now, I'll update `luxonis-ml` so it also installs all optional filesystem packages when using with the `[all]` specifier.

luxonis_train/core/archiver.py
```diff
@@ -0,0 +1,396 @@
...
+
+        head_outputs = {}
+        if head_name == "ClassificationHead":
+            head_outputs["predictions"] = self.outputs[0]["name"]  # TODO: test
+        elif head_name == "EfficientBBoxHead":
+            head_outputs["yolo_outputs"] = [
+                output_dict["name"] for output_dict in self.outputs
+            ]  # TODO: test
+        elif head_name == "ObjectDetectionSSD":
+            raise NotImplementedError  # TODO: boxes, scores
```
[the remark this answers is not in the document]
conorsim: I also confirmed with @tersekmatija by the way that we don't have plans of supporting MobileNet SSD in luxonis-train

tests/unittests/test_core/test_archiver.py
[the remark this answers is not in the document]
jkbmrz: added in commit [5e59c3a](https://github.com/luxonis/luxonis-train/pull/17/commits/5e59c3aebb8a17543446b144697f2e47bc0b67af). For now, all tests are based on ClassificationModel. Later, we can extend tests for all other models.

jkbmrz:
Based on the discussion above, two things remain unresolved:

1. Determining `is_softmax` parameter. As suggested by @conorism we could go for:
> 1/ Hardcoded for each architecture/head in this repo or
> 2/ An option in the training config which we can pull
I’d go for the option 1 for now as we could simply implement it by adding a simple enum like: 
```
class ImplementedHeadsSoxtmaxed(Enum):
  ClassificationHead = True/False
  EfficientBBoxHead = None
  ImplicitKeypointBBoxHead = None
  SegmentationHead = True/False
  BiSeNetHead = True/False
```

2. Expanding tests. As wrote @conorsim:
> It will be nice to also add tests for each type of head in luxonis-train once other questions are resolved.
If I understand correctly, tests should generate a model for each of existing head (ClassificationHead, SegmentationHead, BiSeNetHead, EfficientBBoxHead, ImplicitKeypointBBoxHead) and test if head parameters (also `head_outputs`)  are set correctly?

conorsim:
> If I understand correctly, tests should generate a model for each of the existing heads (ClassificationHead, SegmentationHead, BiSeNetHead, EfficientBBoxHead, ImplicitKeypointBBoxHead) and test if head parameters (also head_outputs) are set correctly?

Yeah this is what I meant.

conorsim (approved):

kozlov721 (approved):
LGTM

PR #18, opened 2024-03-22
Extend NN Archive Generation Test Coverage

jkbmrz:
Extending test coverage of NN archive generation for each type of head in luxonis-train. For each head, we now have the following tests implemented:

- Test if NN archive was created using xz compression (should be the default option)
- Test if NN archive consists of config.json and model.onnx.
- Test if archive ONNX model is valid.
- Test if archived config inputs and outputs are valid.
- Test if archived config head outputs are valid.

### DIFF
--- a/tests/unittests/test_core/test_archiver.py
+++ b/tests/unittests/test_core/test_archiver.py
@@ -9,21 +9,26 @@ import cv2
 import lightning.pytorch as pl
 import numpy as np
 import onnx
-import yaml
 from luxonis_ml.data import LuxonisDataset
+from luxonis_ml.nn_archive.config_building_blocks.base_models import head_outputs
+from parameterized import parameterized
 
 import luxonis_train
 from luxonis_train.core import Archiver
 from luxonis_train.core.exporter import Exporter
 from luxonis_train.core.trainer import Trainer
+from luxonis_train.nodes.enums.head_categorization import ImplementedHeads
 from luxonis_train.utils.config import Config
+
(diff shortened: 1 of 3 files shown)


tests/unittests/test_core/test_archiver.py
```diff
@@ -32,127 +37,323 @@ def setup_class(cls):
...
+            archiver.reset_logging()
+
+            # clear the loaded config instance
+            Config.clear_instance()
 
-        def classification_dataset_generator():
-            for i in range(10):
-                img = np.random.randint(0, 256, (10, 10, 3), dtype=np.uint8)
-                img_file_path = os.path.join(cls.tmp_path, "images", f"img{i}.png")
+    def _make_dummy_ldf(
```
conorsim, line 105: Not important and just an idea, but I could see this being useful across other tests too. We could make a `LuxonisDatasetFactory` as a shared factory which can be used in other tests as well.

conorsim: It might also make sense to instead implement it in `luxonis-ml` and import it here since it would be useful for `luxonis-ml` tests

conorsim (approved):
Thanks - LGTM

kozlov721 (approved):
LGTM

tests/unittests/test_core/test_archiver.py
jkbmrz: Sure, it makes sense to move it to luxonis-ml!

PR #21, opened 2024-04-11
SIGTERM Handling

kozlov721:
Added handling of the SIGTERM signal. The current state of the training is saved and can be later resumed using the `--resume` flag.

### DIFF
--- a/luxonis_train/core/trainer.py
+++ b/luxonis_train/core/trainer.py
@@ -1,3 +1,5 @@
+import os.path as osp
+import signal
 import threading
 from logging import getLogger
 from typing import Any, Literal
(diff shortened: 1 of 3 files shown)


luxonis_train/core/trainer.py
```diff
@@ -56,7 +90,7 @@ def _upload_logs(self) -> None:
 
     def _trainer_fit(self, *args, **kwargs):
         try:
-            self.pl_trainer.fit(*args, **kwargs)
+            self.pl_trainer.fit(*args, ckpt_path=self.resume, **kwargs)
```
conorsim, line 93: Unimportant for this PR, but when we resume from checkpoint, should we consider also resuming the state of some hyperparameters? I'm not sure if that is already abstracted and/or a feature of pytorch lightning.

conorsim (approved):

luxonis_train/core/trainer.py
tersekmatija: Yes, we should probably store all states (optimizer, scheduler, ...) so we can truly continue with the training.

kozlov721: Yeah, pytorch lightning does this automatically.

PR #22, opened 2024-04-21
Task Label Groups Support

kozlov721:
Added support for task label groups as implemented in `luxonis-ml` in [101](https://github.com/luxonis/luxonis-ml/pull/101).

### DIFF
--- a/luxonis_train/models/luxonis_model.py
+++ b/luxonis_train/models/luxonis_model.py
@@ -262,7 +262,7 @@ class LuxonisModel(pl.LightningModule):
 
         @type inputs: L{Tensor}
         @param inputs: Input tensor.
-        @type task_labels: L{Labels} | None
+        @type task_labels: L{TaskLabels} | None
         @param task_labels: Labels dictionary. Defaults to C{None}.
         @type images: L{Tensor} | None
         @param images: Canvas tensor for visualizers. Defaults to C{None}.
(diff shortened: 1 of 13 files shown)


luxonis_train/models/luxonis_model.py
```diff
@@ -259,8 +262,8 @@ def forward(
 
         @type inputs: L{Tensor}
         @param inputs: Input tensor.
-        @type labels: L{Labels} | None
-        @param labels: Labels dictionary. Defaults to C{None}.
+        @type task_labels: L{Labels} | None
```
conorsim, line 265:
```suggestion
        @type task_labels: L{TaskLabels} | None
```
I think this type just needs an update

luxonis_train/utils/boxutils.py
```diff
@@ -425,7 +426,8 @@ def anchors_from_dataset(
 
     widths = []
     inputs = None
-    for inp, labels in loader:
+    for inp, task_labels in loader:
+        labels = next(iter(task_labels.values()))  # TODO: handle multiple tasks
```
conorsim, line 430: > `# TODO: handle multiple tasks`

What is needed for this TODO?

conorsim (commented):
Leaving comments for now in case multi-task support is still needed

conorsim (approved):

PR #23, opened 2024-05-07
Deterministic Training Support

klemen1999:
Added option for full reproducibility of training by specifying the seed.
This also includes deterministic order of creating `Nodes` and using CPU for drawing segmentation masks for visualizations (related to this [torchvision bug](https://github.com/pytorch/pytorch/issues/104217#issuecomment-1612967519))


luxonis_train/core/core.py
```diff
@@ -90,6 +90,11 @@ def __init__(
         # NOTE: overriding logger in pl so it uses our logger to log device info
         rank_zero_module.log = logger
 
+        deterministic = False
+        if self.cfg.trainer.seed is not None:
+            pl.seed_everything(self.cfg.trainer.seed, workers=True)
+            deterministic = True
+
         self.train_augmentations = TrainAugmentations(
```
tersekmatija, line 98: I understand TrainAugmentations will be deterministic due to pl.seed_everything correct?

tersekmatija (approved):
Looks good, thanks!

luxonis_train/core/core.py
klemen1999: Yes, this sets python's random seed, numpy seed and torch seed ([here](https://pytorch-lightning.readthedocs.io/en/1.7.7/_modules/pytorch_lightning/utilities/seed.html#seed_everything))

PR #26, opened 2024-05-13
Updated Tuner

klemen1999:
- Updated Optuna requrements to a newer version compatible with pytorch-lightning>=2.0
- Added `continue_existing_study: bool = True` parameter to the config. If this is set to True (default) then it will use existing study with that name otherwise if name already exists and this is set to False it will throw an error.
- Added logging to Tuner which gives best study parameters at the end


tersekmatija (approved):
Let's see if we can nest better than recursively in MLFlow, otherwise looks good.

kozlov721 (approved):
LGTM

klemen1999:
Now tuning runs are nested together under common parent experiment and can be more easily compared in MLFLow. Parent tracker is only used to report hyperparameters that are currently the best in the whole study.  CC: @tersekmatija 
image
image

luxonis_train/core/tuner.py
```diff
@@ -30,8 +33,23 @@ def __init__(self, cfg: str | dict, args: list[str] | tuple[str, ...] | None):
...
+        cfg_tracker = self.cfg.tracker
+        tracker_params = cfg_tracker.model_dump()
+        self.parent_tracker = LuxonisTrackerPL(
+            rank=rank,
+            mlflow_tracking_uri=self.cfg.ENVIRON.MLFLOW_TRACKING_URI,
+            is_sweep=False,
+            **tracker_params,
+        )
+        if self.parent_tracker.is_mlflow:
+            run = self.parent_tracker.experiment["mlflow"].active_run()
```
tersekmatija, line 47: Will parent run remain active long enough for this to not cause any issues?

luxonis_train/core/tuner.py
```diff
@@ -30,8 +33,23 @@ def __init__(self, cfg: str | dict, args: list[str] | tuple[str, ...] | None):
...
+            self.parent_run_id = run.info.run_id
```
tersekmatija, line 48: Is this used somewhere? What is this set to if MLFlow is not used?

klemen1999: If MLFlow is not used then still this first tracker will report the best hyperparameters but nesting won't be done because it is not supported by e.g. Tensorboard. So every trial will have it's own tensorboard logs with no connection between them. That's why when tuning using MLFlow is recommended.
The whole part in this if block seems like it is not actually needed but if I remove it then nesting is not doen for some reason. Looking into it and I'll update you

tersekmatija: Ok sounds good. Not critical. I think looks good, maybe we just add a comment in code to note this is required for MLFlow to work correctly, then we can merge.

klemen1999: I figured it out: The way tracker is written it creates the actual mlflow run once it is interacted with. And since in our case we need parent tracker to be created first (to then nest all children runs to that one) we need to kep this line in the code. I added a note to the code.
CC: @kozlov721 I think now we can merge

tersekmatija: Is this specific to "mlflow" then, or should this be present for other underlying trackers as well? @klemen1999

klemen1999: It's the case for all trackers. LuxonisTracker is sctructured very similar to other default PL loggers (e.g. [MLFlowLogger](https://lightning.ai/docs/pytorch/stable/_modules/lightning/pytorch/loggers/mlflow.html#MLFlowLogger)) where the actual objects are created when `.experiment` property is first called. This ensures that LuxonisTracker is nicely itegrated with PL but we have to be mindfull of that when using it like this. Although I don't think there will be many cases where one would create just a placeholder tracker, like we are doing for this specific case, so it shouldn't cause to many problems.

tersekmatija: So should we always call this instead of only on:
`if self.parent_tracker.is_mlflow`

klemen1999: `mlflow` is the only tracker that this nesting applies to so that is why I'm doing it just for this.

tersekmatija: Sounds good, I think we can proceed with the merge in this case.

klemen1999:
I've added graceful stop for `LuxonisTrackerPL` with function `finalize()` - this is automatically called when PL process is finished. This should remove the need for explicit `mlflow.end_run` in #30 @kozlov721

PR #27, opened 2024-05-15
Custom Loaders Support

kozlov721:
Added support for using custom loaders.

## Example

**source.py**
```python
from luxonis_train.utils.loaders import BaseLoaderTorch

class CustomLoader(BaseLoaderTorch):
    import torch
    from luxonis_ml.data import LabelType

    from luxonis_train.utils.loaders import LuxonisLoaderTorchOutput

    @property
    def input_shape(self) -> torch.Size:
        return self.torch.Size((1, 3, 256, 256))

    def __getitem__(self, _: int) -> LuxonisLoaderTorchOutput:
        from luxonis_ml.data import LabelType

        img = self.torch.rand(3, 256, 256)
        label = self.torch.zeros(1, 256, 256)
        label[0, 100:200, 100:200] = 1
        labels = {
            "default": {
                LabelType.SEGMENTATION: label,
            }
        }
        return img, labels

    def __len__(self) -> int:
        return 32

    def get_classes(self) -> dict[LabelType, list[str]]:
        return {self.LabelType.SEGMENTATION: ["square"]}
```

```bash
luxonis_train --source source.py \
              --config configs/segmentation_model.yaml \ 
              loader.name CustomLoader loader.params '{}'
```

## Breaking Config Changes

`dataset` section renamed to `loader`. Custom loader can be referenced by its name in the `LOADERS` registry. The default loader is `LuxonisLoaderTorch`. Former keys of the `dataset` field now go under `params` field in the `loader`. 

```yaml
loader:
  name: LuxonisLoaderTorch  # this is the default

  train_view: train
  val_view: val
  test_view: test

  params:
    dataset_name: coco_test
    bucket_storage: gcs

```

### DIFF
--- a/luxonis_train/core/core.py
+++ b/luxonis_train/core/core.py
@@ -7,14 +7,16 @@ import lightning.pytorch as pl
 import lightning_utilities.core.rank_zero as rank_zero_module
 import rich.traceback
 import torch
+import torch.utils.data as torch_data
 from lightning.pytorch.utilities import rank_zero_only  # type: ignore
-from luxonis_ml.data import LuxonisDataset, TrainAugmentations, ValAugmentations
+from luxonis_ml.data import TrainAugmentations, ValAugmentations
 from luxonis_ml.utils import reset_logging, setup_logging
 
 from luxonis_train.callbacks import LuxonisProgressBar
 from luxonis_train.utils.config import Config
 from luxonis_train.utils.general import DatasetMetadata
-from luxonis_train.utils.loaders import LuxonisLoaderTorch, collate_fn
+from luxonis_train.utils.loaders import collate_fn
+from luxonis_train.utils.registry import DATASETS, LOADERS
 from luxonis_train.utils.tracker import LuxonisTrackerPL
 
 logger = getLogger(__name__)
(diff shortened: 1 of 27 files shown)


tests/unittests/test_core/test_archiver.py
```diff
@@ -23,7 +25,8 @@
 HEAD_NAMES = [head_name for head_name in ImplementedHeads.__members__]
 
 
-class TestArchiver:
+@pytest.mark.skip()
```
conorsim, line 28: Why do we skip this test?

conorsim (approved):
Looks like a nice improvement, thanks!

tests/unittests/test_core/test_archiver.py
kozlov721: The NN Archive in the latest dev of luxonis-ml is not compatible with the tests for now (we need to figure out the yolov8 keypoint outputs in the archive)

luxonis_train/core/core.py
```diff
@@ -175,21 +154,26 @@ def __init__(
...
-            batch_size=self.cfg.trainer.batch_size,
-            num_workers=self.cfg.trainer.num_workers,
-            collate_fn=collate_fn,
-            drop_last=self.cfg.trainer.skip_last_batch,
-            sampler=sampler,
-        )
+                sampler = torch_data.WeightedRandomSampler(weights, num_samples)
+
+        self.pytorch_loaders = {
+            view: torch_data.DataLoader(
```
tersekmatija, line 160: What if self.loaders[view] does not exist?

tersekmatija (commented):
1/ Does this affect any of our current online datasets?
2/ Have we tested the kpts training?

Otherwise looks good, thanks!

kozlov721:
> 1/ Does this affect any of our current online datasets?

No, this shouldn't change anything regarding the underlying datasets

>2/ Have we tested the kpts training?

Yes

CaptainTrojan:
@kozlov721 What is the correct way of providing `BucketType` and `BucketStorage` to the default `LuxonisLoaderTorch`? The `DatasetConfig` ensured that strings like `internal`, `local, `gcs` etc. were mapped to the corresponding enum type.

PR #29, opened 2024-05-21
GPUStatsMonitor

kozlov721:
- Added a new callback `GPUStatsMonitor` for monitoring the GPU. 
- The `DeviceStatsMonitor` and `GPUStatsMonitor` are now always added to the trainer

### DIFF
--- a/luxonis_train/callbacks/gpu_stats_monitor.py
+++ b/luxonis_train/callbacks/gpu_stats_monitor.py
@@ -251,7 +251,9 @@ class GPUStatsMonitor(pl.Callback):
         logs = {}
         for i, device_id in enumerate(device_ids):
             for j, (x, unit) in enumerate(keys):
-                logs[f"GPU {device_id}/{x} ({unit})"] = stats[i][j]
+                if unit == "%":
+                    unit = "percent"
+                logs[f"GPU {device_id}/{x} {unit}"] = stats[i][j]
         return logs
 
     def _get_gpu_stat_keys(self) -> List[Tuple[str, str]]:
--- a/luxonis_train/callbacks/gpu_stats_monitor.py
+++ b/luxonis_train/callbacks/gpu_stats_monitor.py
@@ -251,7 +251,7 @@ class GPUStatsMonitor(pl.Callback):
         logs = {}
         for i, device_id in enumerate(device_ids):
             for j, (x, unit) in enumerate(keys):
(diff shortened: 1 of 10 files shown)


luxonis_train/models/luxonis_model.py
```diff
@@ -620,18 +622,23 @@ def configure_callbacks(self) -> list[pl.Callback]:
...
                 dirpath=self.min_val_loss_checkpoints_path,
                 filename=f"{model_name}_loss={{val/loss:.4f}}_{{epoch:02d}}",
                 auto_insert_metric_name=False,
                 save_top_k=self.cfg.trainer.save_top_k,
                 mode="min",
-            )
-        )
+            ),
+        ]
+        if "DeviceStatsMonitor" not in user_callbacks:
```
tersekmatija, line 637: Should we have those enabled by default?

luxonis_train/callbacks/gpu_stats_monitor.py
```diff
@@ -0,0 +1,287 @@
+# Copyright The PyTorch Lightning team.
```
tersekmatija, line 1: I know lightning is currently a requirements anywhere, but should we explicitly mention this in the README as well?
Is there a known reason why this was removed in lightning 2 btw?

luxonis_train/models/luxonis_model.py
kozlov721: They are. But the user can still define them themselves in which case they override the default.

luxonis_train/callbacks/gpu_stats_monitor.py
kozlov721: >I know lightning is currently a requirements anywhere, but should we explicitly mention this in the README as well?

Yeah we can mention that `luxonis-train` is build on top of lightning. 

>Is there a known reason why this was removed in lightning 2 btw?

It wasn't removed entirely; it was replaced by the `DeviceStatsMonitor`, but the `DeviceStatsMonitor` doesn't use `nvidia-smi`, only torch memory stats, which don't include the overall gpu utilization. There's some discussion about it in this [issue](https://github.com/Lightning-AI/pytorch-lightning/issues/11796), but there's been no new activity since 2022.

tersekmatija: > Yeah we can mention that luxonis-train is build on top of lightning.

Ok, let's mention it. Sounds good on the rest.

conorsim (approved):

PR #30, opened 2024-05-21
Graceful Stop MLFlow Run

kozlov721:
Added `mlflow.end_run` to the graceful exit.

### DIFF
--- a/luxonis_train/core/trainer.py
+++ b/luxonis_train/core/trainer.py
@@ -63,6 +63,9 @@ class Trainer(Core):
             self._upload_logs()
 
             if self.cfg.tracker.is_mlflow:
+                import mlflow
+                from mlflow.entities.run_status import RunStatus
+
                 logger.info("Uploading checkpoint to MLFlow.")
                 fs = LuxonisFileSystem(
                     "mlflow://",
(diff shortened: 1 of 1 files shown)


luxonis_train/core/trainer.py
```diff
@@ -74,6 +77,7 @@ def graceful_exit(signum: int, _):
                     remote_path="resume.ckpt",
                     mlflow_instance=self.tracker.experiment.get("mlflow", None),
                 )
+                mlflow.end_run(RunStatus.to_string(RunStatus.FAILED))
```
tersekmatija, line 80: We use two trackers in tuner with optuna. How does this work in such case? CC @klemen1999

luxonis_train/core/trainer.py
```diff
@@ -63,6 +63,9 @@ def graceful_exit(signum: int, _):
             self._upload_logs()
 
             if self.cfg.tracker.is_mlflow:
```
tersekmatija, line 65: What if TensorBoard or WandB would be used? Does it handle transition to fail automatically?

luxonis_train/core/trainer.py
klemen1999: I think the ending of tracker runs should be handled inside `LuxonisTrackerPL` since base PL Logger has `finalize` method that is called on every teardown with status. I'll add this to the #26 and you should be then able to remove this line from here. 
Regarding 2 trackers in Tuner: MLFlow uses stack for active trackers so it would end them in reverse order that they are started which works fine

tersekmatija: Ok, sounds good. Let's add to #26. And we can then remove @kozlov721 here.

PR #32, opened 2024-05-22
Added active param to augmentations

klemen1999:
- Added `active` parameter for each augmentation
- [Augmentation](https://github.com/luxonis/luxonis-ml/blob/06dedb7bb0e7caaee1663e815925456a51c46595/luxonis_ml/data/augmentations/utils.py#L16) gets only currently active augmentations
- This makes it possible for `Tuner` to turn on/off specific augmentation per trial. Example part of config:
```yaml
trainer:
  ...
  preprocessing:
    ...
     augmentations:
       - name: Defocus
         params:
           p: 0.1
tuner:
  params:
    trainer.preprocessing.augmentations.0.active_categorical: [True, False]
```

### DIFF
--- a/luxonis_train/__main__.py
+++ b/luxonis_train/__main__.py
@@ -137,7 +137,9 @@ def inspect(
 
     augmentations = (TrainAugmentations if view == "train" else ValAugmentations)(
         image_size=image_size,
-        augmentations=[i.model_dump() for i in cfg.trainer.preprocessing.augmentations],
+        augmentations=[
+            i.model_dump() for i in cfg.trainer.preprocessing.get_active_augmentations()
+        ],
         train_rgb=cfg.trainer.preprocessing.train_rgb,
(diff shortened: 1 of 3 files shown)


klemen1999:
Nice future improvement would be to make `_subset` keyword for `Tuner` (e.g. trainer.preprocessing.augmentations: [["Defocus", "Sharpen", "RandomRotate90"], 2] - would make subset of size 2 from the list) but we would have to think how to make it general enough. If we make it for augmentations only this would mean just turning off `active` parameter for all unselected augs but if for example user tries to use it on parameter that doesn't have this `active` flag it would fail. Essentially the question is how general do we want to make this "_subset" tunning.

kozlov721 (approved):
LGTM

tersekmatija (approved):

conorsim (approved):

PR #36, opened 2024-06-01
Multi Input Support

CaptainTrojan:
# Motivation
The goal is to support multi-input models and loaders in luxonis-train. 

# Examples
## Multi-input loader

The loader simulates an input consisting of stereo images (left and right), disparity maps, and point clouds. It also identifies "left" as the primary image type, which can be used in visualizers.
```python
class CustomMultiInputLoader(BaseLoaderTorch):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)

    @property
    def input_shape(self) -> dict[str, Size]:
        return {
            "left": torch.Size([3, 224, 224]),
            "right": torch.Size([3, 224, 224]),
            "disparity": torch.Size([1, 224, 224]),
            "pointcloud": torch.Size([1000, 3]),
        }

    @property
    def image_source(self) -> str:
        return "left"

    def __getitem__(self, idx):
        # Fake data
        left = torch.rand(3, 224, 224, dtype=torch.float32)
        right = torch.rand(3, 224, 224, dtype=torch.float32)
        disparity = torch.rand(1, 224, 224, dtype=torch.float32)
        pointcloud = torch.rand(1000, 3, dtype=torch.float32)
        inputs = {
            "left": left,
            "right": right,
            "disparity": disparity,
            "pointcloud": pointcloud,
        }

        # Fake labels
        segmap = torch.zeros(1, 224, 224, dtype=torch.float32)
        labels = {
            "segmentation": (segmap, LabelType.SEGMENTATION)
        }

        return inputs, labels

    def __len__(self):
        return 10

    def get_classes(self) -> dict[LabelType, list[str]]:
        return {LabelType.SEGMENTATION: ["square"]}
```
## Multi-input model

This model includes nodes such as FullBackbone, RGBDBackbone (processing left, right, and disparity inputs), and PointcloudBackbone (processing pointcloud inputs). It also features FusionNeck and FusionNeck2 for combining outputs from these backbones, and CustomSegHead1 and CustomSegHead2 which also make use of the different fusion necks and inputs. The model showcases our capability to handle intricate data fusion, multi-modal processing, and serving loader data to any node in the graph.
```yaml
model:
  name: example_multi_input
  nodes:
    - name: FullBackbone
      alias: full_backbone  # By default loads all loader sub-elements since no inputs nor input_sources are provided

    - name: RGBDBackbone
      alias: rgbd_backbone
      input_sources:
        - left
        - right
        - disparity

    - name: PointcloudBackbone
      alias: pointcloud_backbone
      input_sources:
        - pointcloud

    - name: FusionNeck
      alias: fusion_neck
      inputs:
        - rgbd_backbone
        - pointcloud_backbone
      input_sources:
        - disparity

    - name: FusionNeck2
      alias: fusion_neck_2
      inputs:
        - rgbd_backbone
        - pointcloud_backbone
        - full_backbone
    
    - name: CustomSegHead1
      alias: head_1
      inputs:
        - fusion_neck

    - name: CustomSegHead2
      alias: head_2
      inputs:
        - fusion_neck
        - fusion_neck_2
      input_sources:
        - disparity
```
# Test changes
[Updated collate_fn test ](https://github.com/luxonis/luxonis-train/pull/36/commits/6809f98e4f1b39b865175d2efed2c2cee87b44ea)due to how inputs were changed from `Tensor`s to `dict[str, Tensor]`.

# Breaking changes
Changed the type of `original_in_shape` of `LuxonisModel` from `Size` to `dict[str, Size]` (multiple inputs), so any custom models making use of this will stop working. To resolve this, users can make use of the new structure [like this](https://github.com/luxonis/luxonis-train/commit/0e1ceffb5ea35bfc9a201da063d1b322e894bc78).

### DIFF
--- a/luxonis_train/utils/loaders/base_loader.py
+++ b/luxonis_train/utils/loaders/base_loader.py
@@ -9,8 +9,9 @@ from torch.utils.data import Dataset
 from luxonis_train.utils.registry import LOADERS
 from luxonis_train.utils.types import Labels, LabelType
 
-LuxonisLoaderTorchOutput = tuple[Tensor, dict[str, Labels]]
-"""LuxonisLoaderTorchOutput is a tuple of images and corresponding labels."""
+LuxonisLoaderTorchOutput = tuple[dict[str, Tensor], dict[str, Labels]]
+"""LuxonisLoaderTorchOutput are two dictionaries, the first one contains the input data
+and the second one contains the labels."""
 
 
 class BaseLoaderTorch(
@@ -27,14 +28,55 @@ class BaseLoaderTorch(
         self,
         view: str,
         augmentations: Augmentations | None = None,
+        images_name: str | None = None,
     ):
         self.view = view
         self.augmentations = augmentations
+        self._images_name = images_name
+
+    @property
+    def images_name(self) -> str:
+        """Name of the input image group.
+
+        Example: 'features'
+        """
+        return self._images_name
 
     @property
     @abstractmethod
-    def input_shape(self) -> Size:
-        """Input shape in [N,C,H,W] format."""
+    def input_shape(self) -> dict[str, Size]:
+        """
+        Shape of each loader group (sub-element), WITHOUT batch dimension.
+        Examples:
+
+        1. Single image input:
+            {
+                'image': torch.Size([3, 224, 224]),
+            }
+
+        2. Image and segmentation input:
+            {
+                'image': torch.Size([3, 224, 224]),
+                'segmentation': torch.Size([1, 224, 224]),
+            }
+
+        3. Left image, right image and disparity input:
+            {
+                'left': torch.Size([3, 224, 224]),
+                'right': torch.Size([3, 224, 224]),
+                'disparity': torch.Size([1, 224, 224]),
+            }
+
+        4. Image, keypoints, and point cloud input:
+            {
+                'image': torch.Size([3, 224, 224]),
+                'keypoints': torch.Size([17, 2]),
+                'point_cloud': torch.Size([20000, 3]),
+            }
+
+        @rtype: dict[str, Size]
+        @return: A dictionary mapping group names to their shapes.
+        """
         ...
 
     @abstractmethod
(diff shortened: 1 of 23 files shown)


CaptainTrojan:
I see that the MacOS tests fail for all open PRs, so opening this one as well.

tests/integration/test_multi_input.py
```diff
@@ -0,0 +1,195 @@
...
+    # opts = [
+    #     "trainer.epochs",
+    #     "3",
+    #     "trainer.validation_interval",
+    #     "3",
+    # ]
+
+    Trainer(f"configs/{config_file}").train()
+
+    # TODO add export and eval tests
```
kozlov721, line 181: Does the export and eval not work yet?

luxonis_train/utils/loaders/base_loader.py
```diff
@@ -27,14 +28,55 @@ def __init__(
...
         view: str,
         augmentations: Augmentations | None = None,
+        images_name: str | None = None,
     ):
         self.view = view
         self.augmentations = augmentations
+        self._images_name = images_name
+
+    @property
+    def images_name(self) -> str:
```
kozlov721, line 38: I would rename this to maybe `source_names`. It doesn't have to necessarily be images

kozlov721: Also raising an error in case it's `None`.

kozlov721: >I would rename this to maybe source_name. It doesn't have to necessarily be images

Or do we always expect this to be an image?

tests/integration/test_multi_input.py
CaptainTrojan: Fixed, I forgot to add the remaining tests.

luxonis_train/utils/loaders/base_loader.py
CaptainTrojan: The property `images_name` I created due to compatibility with image loggers/visualizers, which expect some of the loader inputs to contain image-like objects which it can process in OpenCV, log as images, etc. Right now, the `config.yaml` owner is responsible for setting this property for their custom multi-input logger so that the rest of the code knows which inputs are images.

CaptainTrojan: > Also raising an error in case it's None.
I [moved](https://github.com/luxonis/luxonis-train/commit/1cdcb9f29a365558a74df3c410b1bcf6ce82556d) the declaration of `images_name` to the config, which has `'features'` as the value by default, so unless the user implements their custom loader overriding this property as `None`, it should never be `None`. Is that how you produced this scenario?

kozlov721: I see, that makes sense. Though I'd maybe rename it to `image_source` or something like that.

kozlov721:
> I see that the MacOS tests fail for all open PRs, so opening this one as well.

Yeah, you can ignore the MacOS tests. The MacOS machines are running out of memory, I'm planning to fix it eventually, but it's low priority.

PR #37, opened 2024-06-02
Support for LuxonisML - Annotation Refactor

kozlov721:
- Support for refactored annotations in LuxonisML ([110](https://github.com/luxonis/luxonis-ml/pull/110))
  - Option to specify `task` field under `node`'s config
  - The specified `task` is used by the node and all the attached modules
- Test case for multi-task model

### DIFF
--- a/configs/resnet_multitask_model.yaml
+++ b/configs/resnet_multitask_model.yaml
@@ -1,0 +1,110 @@
+
+model:
+  name: resnet50_classification
+  nodes:
+    - name: ResNet
+      params:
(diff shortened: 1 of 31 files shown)


conorsim (approved):
The code LGTM - not sure if the ubuntu tests need to be updated though

kozlov721:
>not sure if the ubuntu tests need to be updated though

they are using the `dev` branch of `luxonis-ml` which doesn't yet have the new annotations merged. I'll re-run the tests when it gets merged.

PR #39, opened 2024-06-10
add OCR Decoding support - WIP

N950:
Adding Nodes/Loss/Metric for OCR recognition/decoding
Conformity to contribution guidelines will be fixed before closing


luxonis_train/attached_modules/losses/ocr_decoding_loss.py
```diff
@@ -0,0 +1,49 @@
+import torch
+from torch import Tensor, nn
+
+from .base_loss import BaseLoss
+
+
+class FocalCTC(BaseLoss[Tensor, Tensor]):
```
kozlov721, line 7: Add class attribute `supported_labels = [LabelType.TEXT`]

luxonis_train/nodes/ocr_decoder.py
```diff
@@ -0,0 +1,171 @@
...
+import torch
+import torch.nn as nn
+import torchvision
+from torch import Tensor
+
+from .base_node import BaseNode
+from luxonis_train.utils.types import LabelType
+
+
+class ResBlock(nn.Module):
```
kozlov721, line 17: I believe both `ResBlock` nad `UpSample` should already be implemented in `luxonis_train.nodes.blocks` or in one of the backbones

luxonis_train/nodes/ocr_decoder.py
```diff
@@ -0,0 +1,171 @@
...
+        self.block = nn.Sequential(
+            nn.Conv2d(in_channels=ch_in, out_channels=ch_out, kernel_size=kernel_size, stride=stride, padding=padding)
+        )
+
+    def forward(self, x):
+        out = self.block(x)
+        return out
+
+
+class OCRDecoderBackbone(BaseNode):
```
kozlov721, line 51: Instead of `_task_type` argument, add a class attribute `tasks: list[LabelType] = [LabelType.TEXT]`

luxonis_train/attached_modules/losses/ocr_decoding_loss.py
kozlov721: Also the metrics and losses will probably require custom `prepare` method that takes the packed output of the head and labels and prepares argument to be passed to `forward` or `update`.

PR #42, opened 2024-06-19
Multi-GPU Training Support

MatejRojec:
- Modify the backbone models to support multi-GPU training using the DDP (Distributed Data Parallel) strategy.
- Resolve the `ddp_find_unused_parameters_true` error, which occurs when some backbone components are not utilized in the forward pass.

### DIFF
--- a/luxonis_train/nodes/efficientnet.py
+++ b/luxonis_train/nodes/efficientnet.py
@@ -28,7 +28,9 @@ class EfficientNet(BaseNode[Tensor, list[Tensor]]):
             pretrained=download_weights,
         )
         efficientnet_lite0_model.classifier = nn.Identity()
-        self.out_indices = [1, 4, 6]
+        self.out_indices = [0, 1, 2, 4, 6]
+        efficientnet_lite0_model.bn2 = nn.Identity()
+        efficientnet_lite0_model.conv_head = nn.Identity()
         self.backbone = efficientnet_lite0_model
 
     def forward(self, x: Tensor) -> list[Tensor]:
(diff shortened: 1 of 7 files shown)


kozlov721 (approved):
LGTM

luxonis_train/nodes/efficientnet.py
```diff
@@ -39,4 +40,10 @@ def forward(self, x: Tensor) -> list[Tensor]:
             x = m(x)
             if i in self.out_indices:
                 outs.append(x)
+
+        x = self.backbone.conv_head(x)
+        x = self.backbone.bn2(x)
+        x = self.backbone.act2(x)
+        x = self.backbone.global_pool(x)
```
tersekmatija, line 47: Is there a reason we append flattened features here instead of before global pool?
It seems also we have a list of size 5?

tersekmatija: Whereas elsewhere it's a list of size 4?

tersekmatija (changes requested):

luxonis_train/nodes/efficientnet.py
tersekmatija: I think maybe we check the paper if they define "blocks" to find out which features we should take?

tersekmatija:
@kozlov721 Do we require anywhere downstream that backbone outputs exactly 4 feature maps?

kozlov721:
> @kozlov721 Do we require anywhere downstream that backbone outputs exactly 4 feature maps?

No, I don't think so

tersekmatija:
How would neck accept this then, the last 3 feature maps for example? @kozlov721

kozlov721:
> How would neck accept this then, the last 3 feature maps for example? @kozlov721

The `RepPANNeck` needs at least as many feature maps as the value of `num_heads` parameter (3 by default). Then it uses the feature maps from the last one in the list.

tersekmatija (approved):

tersekmatija:
Okay, I think we can merge this then.

PR #49, opened 2024-07-30
Fixed New `mlflow` Run

kozlov721:
Fixed a bug introduced in #26 causing logs and checkpoints to be uploaded to a new `mlflow` run instead of the current one.


conorsim (approved):
Looks good and a quick test confirms it works.

PR #51, opened 2024-07-31
Adaptive Detection Loss Speed-Up

JSabadin:
### Performance Improvement

Previously, the preparation phase took **36 minutes** out of a total of **2.3 hours per epoch**. The bottleneck was identified in the calculation at this line: [adaptive_detection_loss.py#L108](https://github.com/luxonis/luxonis-train/blob/main/luxonis_train/attached_modules/losses/adaptive_detection_loss.py#L108), which was executed on every single call of the `prepare` function.

To address this, Imodified the code to calculate this only once. As a result, the speed per epoch on my local PC improved significantly, reducing from **0.13 seconds per call** (measured precisely using `torch.cuda.synchronize()`) to **0.013 seconds per call.** This optimization represents a substantial performance boost!


kozlov721 (approved):
LGTM

conorsim (approved):

JSabadin:
Not yet ready to be merged, as it did not affect the training epoch time. This issue needs to be investigated first

PR #53, opened 2024-08-03
Full Custom Tasks Support

kozlov721:
Added full support for custom task names. When using custom task names, a new field `task` has to be specified in the configuration of the corresponding head node. The `task` gets propagated to all the modules attached to the head (losses, metrics, _etc._).   

## Example

**Labels**
```python
labels = {"segmentation-task": ...,
          "boundingbox-task": ...,
          "keypoints-task": ...}
```

**`config.yaml`**
```yaml
nodes:
    ...
    - name: SegmentationHead
      # single-task head requires just the new task name
      task: segmentation-task

    - name: EfficientKeypointBBoxHead
      # multi-task heads require a mapping from the default names 
      # (string representations of the corresponding `LabelType`) to the new custom task names
      task:
        keypoints: keypoints-task
        boundingbox: boundingbox-task
```

## Subclassing Changes

Minor breaking changes introduced for implementing custom nodes and attached modules.
- Specifying task types and required labels using class attributes instead of `__init__` arguments.
- Removed protocols for attached nodes
  - `BaseAttachedNode.get_label` and `BaseAttachedNode.get_input_tensor` now do the checking insteadd  

### Nodes

**old_node.py**
```python
class CustomNode(BaseNode):
    def __init__(self, **kwargs):
        super().__init__(**kwargs, _task_type=LabelType.SEGMENTATION)
```
**new_node.py**
```python
class CustomNode(BaseNode):
    # either a list of `LabelType`s or a mapping from `LabelType` to a default task name,
    # the following is equivalent to `tasks = {LabelType.SEGMENTATION: "segmentation"}`
    tasks = [LabelType.SEGMENTATION]
    def __init__(self, **kwargs):
        super().__init__(**kwargs)
```

### Attached Modules

**old_loss.py**
```python
class CustomLoss(BaseLoss):
    def __init__(self, **kwargs):
        super().__init__(
            **kwargs, 
            required_labels=LabelType.SEGMENTATION,
            protocol=SegmentationProtocol,
        )
    def prepare(self, inputs: Packet[Tensor], labels: Labels):
        # presence of the "segmentation" key is validated beforehand by the `Protocol`
        predicted_masks = inputs["segmentation"]
        # validated beforehand using `required_labels`
        label_mask = inputs["segmentation"]
```

**new_loss.py**
```python
class CustomLoss(BaseLoss):
    # list of all tasks this module supports
    # can contain either standalone `LabelType`s or tuples of `LabelType`s
    # tuple specifies that all of the tasks in the apple must be provided  
    supported_labels = [LabelType.SEGMENTATION]
    # [(LabelType.BOUNDINGBOX, LabelType.KEYPOINTS)] would mean that the module
    # requires both of these tasks to be present in node outputs and labels
    def __init__(self, **kwargs):
        super().__init__(**kwargs)

    def prepare(self, inputs: Packet[Tensor], labels: Labels):
        # returns the correct input tensor from the node outputs
        predicted_masks = self.get_input_tensor(inputs, LabelType.SEGMENTATION)
        # returns the correct label based on the task of the attached `BaseNode`
        label_mask = self.get_label(labels, LabelType.SEGMENTATION)
```


luxonis_train/nodes/base_node.py
```diff
@@ -62,41 +104,59 @@ class BaseNode(
...
     @param in_sizes: List of input sizes for the node.
         Provide only in case the `input_shapes` were not provided.
+
+    @type _tasks: dict[LabelType, str] | None
+    @param _tasks: Dictionary of tasks that the node supports. Overrides the
+        class L{tasks} attribute. Shouldn't be provided by the user in most cases.
+
+    @type input_protocols: list[type[BaseModel]]
+    @ivar input_protocols: List of input protocols used to validate inputs to the node.
+        Defaults to [L{FeaturesProtocol}]. Protoco
```
klemen1999, line 121: Missing "l" at the end of "Protoco"?

luxonis_train/nodes/base_node.py
```diff
@@ -323,12 +417,16 @@ def wrap(self, output: ForwardOutputT) -> Packet[Tensor]:
                 raise IncompatibleException(
                     "Default `wrap` expects a single tensor or a list of tensors."
                 )
-        return {self._task or "features": outputs}
+        try:
+            task = self.task
```
klemen1999, line 421: Should we update the docstring here to say that it wraps into a Packet with task key if present otherwise uses "features" key?

luxonis_train/nodes/base_node.py
```diff
@@ -341,24 +439,28 @@ def run(self, inputs: list[Packet[Tensor]]) -> Packet[Tensor]:
...
+        for key in list(wrapped.keys()):
+            if key in str_tasks:
+                value = wrapped.pop(key)
+                wrapped[self.get_task_name(LabelType(key))] = value
+        return wrapped
 
     def validate(self, data: list[Packet[Tensor]]) -> list[Packet[Tensor]]:
-        """Validates the inputs against `in_protocols`."""
-        if len(data) != len(self.in_protocols):
+        """Validates the inputs against `inpur_protocols`."""
```
klemen1999, line 451: Typo in docstring "inpur"
```suggestion
        """Validates the inputs against `input_protocols`."""
```

luxonis_train/nodes/README.md
```diff
@@ -24,10 +24,9 @@ arbitrarily as long as the two nodes are compatible with each other.
 
 Every node takes these parameters:
 
-| Key          | Type        | Default value | Description                                                                                                               |
-| ------------ | ----------- | ------------- | ------------------------------------------------------------------------------------------------------------------------- |
-| attach_index | int \| None | None          | Index of previous output that the head attaches to. Each node has a sensible default. Usually should not be manually set. |
-| n_classes    | int \| None | None          | Number of classes in the dataset. Inferred from the dataset if not provided.                                              |
+| Key       | Type        | Default value | Description                                                                  |
+| --------- | ----------- | ------------- | ---------------------------------------------------------------------------- |
```
klemen1999, line 28: Why did we remove "attach_index" from docs?

klemen1999 (approved):

klemen1999 (approved):
LGTM!

Just one question: Now that most of the protocols were removed do we still keep the `FeaturesProtocol`?

kozlov721:
>Now that most of the protocols were removed do we still keep the FeaturesProtocol?

That one is used in `BaseNode` subclasses which still use protocols. I'll probably remove it too at some point, but that would be in a separate PR. 

>Are the LuxonisDataset datasets created before still compatible with this newer version of luxonis-train/luxonis-ml?

No, only datasets created after [#110](https://github.com/luxonis/luxonis-ml/pull/110) are compatible (`luxonis-ml>=0.2.0`).

luxonis_train/nodes/README.md
kozlov721: It was removed from the constructor parameters, but I added back a section to the README describing the overridable class attributes.

conorsim (approved):
I didn't review the code, but from manual testing some custom task applications it looks good

PR #56, opened 2024-08-12
Fix: Update NN Archive Generation

conorsim:
This PR updates the `Archiver` to use the official 1.0 NN archive schema. Specifically, there are updates to include the `parser` field and to use the `metadata` field for the head.

I haven't updated `test_archiver.py` since we're currently [skipping](https://github.com/luxonis/luxonis-train/blob/611b93c92f8bdb899a297a2b3ce486d2bbc4c422/tests/unittests/test_core/test_archiver.py#L27) it. Thoughts on updating it? I'm not sure why it's currently skipped.


kozlov721:
> I haven't updated test_archiver.py since we're currently [skipping](https://github.com/luxonis/luxonis-train/blob/611b93c92f8bdb899a297a2b3ce486d2bbc4c422/tests/unittests/test_core/test_archiver.py#L27) it. Thoughts on updating it? I'm not sure why it's currently skipped.

We're skipping because they were outdated and were causing the checks to fail. It was only temporary until we update the archiver, so if you could update the tests too, that would be great.

kozlov721 (approved):
LGTM

jkbmrz (changes requested):
When defining inputs, we are missing the `layout` parameter (lettercode interpretation of the input data dimensions) → e.g. in [_add_input](https://github.com/luxonis/luxonis-train/blob/4eae901d2ce58873c18d6be8cf6e2f6ac9250aa4/luxonis_train/core/archiver.py#L168) method. Otherwise LGTM.

jkbmrz (approved):

PR #60, opened 2024-08-19
Stricter Config Types

kozlov721:
Changed types in `luxonis_train.utils.config.Config` to be stricter and not allow invalid values to be used.

### DIFF
--- a/luxonis_train/utils/config.py
+++ b/luxonis_train/utils/config.py
@@ -1,6 +1,6 @@
 import logging
 import sys
-from typing import Annotated, Any, Literal
+from typing import Annotated, Any, Literal, TypeAlias
 
 from luxonis_ml.data import LabelType
 from luxonis_ml.utils import (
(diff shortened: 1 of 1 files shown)


luxonis_train/utils/config.py
```diff
@@ -32,32 +44,32 @@ class MetricModuleConfig(AttachedModuleConfig):
...
     include_metrics: bool = True
     include_visualizers: bool = True
+    params: Params = {}
 
 
 class ModelConfig(BaseModelExtraForbid):
     name: str = "model"
     predefined_model: PredefinedModelConfig | None = None
-    weights: str | None = None
+    weights: FilePath | None = None
```
klemen1999, line 72: Does this still accept regular string through yaml or as an override parameter?

luxonis_train/utils/config.py
```diff
@@ -32,32 +44,32 @@ class MetricModuleConfig(AttachedModuleConfig):
 
 class FreezingConfig(BaseModelExtraForbid):
     active: bool = False
-    unfreeze_after: int | float | None = None
+    unfreeze_after: NonNegativeInt | NonNegativeFloat | None = None
```
klemen1999, line 47: How are floats actually handled? E.g. if I set unfreeze_after=2.3 will if actually unfreeze after 2.3 epochs or does it round the number? If the latter then we can just set so we only accept int

klemen1999 (approved):
LGTM, just left some comments

luxonis_train/utils/config.py
kozlov721: Float specifies the percentage of the num epochs. So with 100 epochs and `unfreeze_after` set to 0.2, it would unfreeze after 20 epochs.

PR #63, opened 2024-08-20
Option to Parse Data in `LuxonisLoaderTorch`

kozlov721:
Added support for loading data from a directory instead of using an existing `LuxonisDataset`. 

The following example would create a new `LuxonisDataset` from the dataset directory and save it under the name `"dataset_coco"`. When run for a second time, the dataset will be re-used instead of being generated again.

**`config.yaml:`**
```yaml
loader:
  params:
    dataset_dir: path/to/dataset
    dataset_type: coco
    dataset_name: dataset_coco
    delete_existing: false

```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -153,3 +153,4 @@ mlruns
 wandb
 tests/_data
 tests/integration/_test-output
+data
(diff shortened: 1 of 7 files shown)


klemen1999 (approved):
LGTM

PR #65, opened 2024-08-26
Fail-Proof Checkpoint Usage in Callbacks

kozlov721:
- Callbacks that require an existing model checkpoint are more fail-proof now. 
  - `ArchiveOnTrainEnd` and `ExportOnTrainEnd`
  - Added new parameter `preferred_checkpoint` that can be either `'metric"`, or `"loss"`
    - `"metric"` by default  
    -  If the preferred checkpoint doesn't exist, the other checkpoint type is used. If neither of them exists, the callback is skipped.
- If no metric has `is_main_metric` attribute set to `True`, the first metric is selected to be the main metric.

### DIFF
--- a/luxonis_train/utils/config.py
+++ b/luxonis_train/utils/config.py
@@ -91,7 +91,11 @@ class ModelConfig(BaseModelExtraForbid):
             name = metric.alias or metric.name
             logger.info(f"Setting '{name}' as main metric.")
         else:
-            logger.error("No metrics specified. This is likely unintended.")
+            logger.error(
+                "No metrics specified. "
+                "This is likely unintended unless "
+                "the configuration is not used for training."
+            )
         return self
 
     @model_validator(mode="after")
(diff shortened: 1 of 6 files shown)


luxonis_train/utils/config.py
```diff
@@ -77,6 +77,23 @@ class ModelConfig(BaseModelExtraForbid):
...
+                return self
+
+        logger.warning("No main metric specified.")
+        if self.metrics:
+            metric = self.metrics[0]
+            metric.is_main_metric = True
+            name = metric.alias or metric.name
+            logger.info(f"Setting '{name}' as main metric.")
+        else:
+            logger.error("No metrics specified. This is likely unintended.")
```
klemen1999, line 94: Do we expect any crashes if no metrics is specified or is this handled correctly?

klemen1999 (approved):
LGTM

luxonis_train/utils/config.py
kozlov721: No, it would just mean we can't use checkpoints based on metrics. And it can also be a valid when the config is only for export/archiving etc

kozlov721: But I changed the log message to reflect this valid use case.

PR #69, opened 2024-09-09
Code Cleanup and Improved Tests

kozlov721:
## DevOps
- Synced the CI pipeline with the one in `luxonis-ml`
  - Added automatic labeler and assigner
- Removed MacOS tests because they're running out of memory
- Added type checking requirement
- Set the minimum coverage to 90 %
- Updated information in [CONTRIBUTING.md](https://github.com/luxonis/luxonis-train/blob/cleanup/CONTRIBUTING.md)
- Changed formatter settings to conform to PEP 8

## Cleanup
  - Removed dead code
  - Added missing docs
  - Added missing types
  - Some modules moved and renamed
    - Loaders, assigners, callbacks and schedulers moved from `luxonis_train.utils` to `luxonis_train`
    - `luxonis_train.utils` better organized 
  - Changed types of some raised exceptions 
  - Improved error messages
 
## Tests
- Increased coverage to 97 %
- Added sanity tests for every backbone/head combination
  - Groundwork for overfit tests, can be turned on by specifying `LUXONIS_TRAIN_OVERFIT` env variable, but not yet reliable

## Fixes
- Fixed `SmoothBCEWithLogitsLoss`
- Fixed `SoftmaxFocalLoss`
  - Removed incorrect multiclass implementation
  - Currently both focal losses only support binary tasks, support for multiclass will be added later

## Changes
- Added `GradientAccumulationScheduler`, `StochasticWeightAveraging`, `Timer`, and `ModelPruning` callbacks to the callback REGISTRY
- added `trainer.deterministic` option to the config file as a way to use `"warn"` option for models incompatible with determinism
- added `size_multiplier` option to `luxonis_train inspect` CLI command to better display large images
- removed the `Protocol` pattern from `BaseNode`
- added F1 score to `MeanAveragePrecision` for bounding boxes

## Discovered Issues
- The `ImplicitKeypointBBoxHead` does not train
  - Wasn't able to find out why
  - Will address in a separate PR
  - For now, changed the default keypoint head in the predefined keypoint model to `EfficientKeypointBBoxHead` 
    - That one was more performant anyways as far as I know


luxonis_train/nodes/heads/efficient_bbox_head.py
```diff
@@ -27,32 +25,30 @@ class EfficientBBoxHead(
...
-
         @type max_det: int
-        @param max_det: Maximum number of detections retained after NMS. Defaults to C{300}.
+        @param max_det: Maximum number of detections retained after NMS. Defaults to
+            C{300}.
         """
         super().__init__(**kwargs)
 
-        self.n_heads = n_heads
+        self.num_heads = num_heads
```
klemen1999, line 51: Should we rather stick with `n_heads` as we have `n_classes`, `n_keypoints`,... Same question for other heads (efficient_keypoint and implicit_keypoint), modules and docks where `num_heads` appears

klemen1999 (commented):
Generally LGTM, just left some comments we can discuss further before merging

luxonis_train/attached_modules/losses/implicit_keypoint_bbox_loss.py
```diff
@@ -153,19 +162,20 @@ def prepare(
...
 
-        kpts = self.get_label(labels, LabelType.KEYPOINTS)[0]
-        boxes = self.get_label(labels, LabelType.BOUNDINGBOX)[0]
+        kpt_label = self.get_label(labels, LabelType.KEYPOINTS)
+        bbox_label = self.get_label(labels, LabelType.BOUNDINGBOX)
 
-        nkpts = (kpts.shape[1] - 2) // 3
-        targets = torch.zeros((len(boxes), nkpts * 3 + self.box_offset + 1))
-        targets[:, :2] = boxes[:, :2]
+        targets = torch.zeros(
```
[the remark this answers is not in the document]
kozlov721: I'll double check, but the implicit keypoint head doesn't work on `dev` either, so it has been broken for some time already.

luxonis_train/__init__.py
```diff
@@ -1,7 +1,8 @@
 from .attached_modules import *
 from .core import *
+from .loaders import *
 from .models import *
 from .nodes import *
+from .optimizers import *
+from .schedulers import *
 from .utils import *
-
-__version__ = "0.0.1"
```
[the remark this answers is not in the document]
kozlov721: It was already in `pyproject.toml` and I thought `__version__` gets populated automatically, but I see that's not the case. Fixed it by pointing the `pyproject.toml` version to this instead to avoid the duplicate version info.

luxonis_train/nodes/heads/efficient_bbox_head.py
kozlov721: Should we rename all the `num_something` to `n_something` to be consistent? Like `n_workers`, `n_logged_images` etc.?

klemen1999: IMO we should yeah

PR #70, opened 2024-09-10
DDRNet for Semantic Segmentation

sokovninn:
This PR integrates the DDRNet model to enhance semantic segmentation performance.

## Comparison with Current Model (MicroNet + Segmentation Head)
Parameters: DDRNet has 5.7M, compared to 2.1M for MicroNet.
Training Time: Both models have similar single-epoch training times.
## Metrics After 1 Epoch
MicroNet + Segmentation Head:

- Jaccard Index: 0.00873
- F1 Score: 0.69811

DDRNet-23-slim:

- Jaccard Index: 0.00952
- F1 Score: 0.70069

## Possible improvements

- Compute Jaccard Index every n epochs to increase training speed.
- Improve handling of auxiliary heads for export.
- Use pretrained weights.
- Add Online Hard Example Mining (OHEM) with Cross Entropy Loss.
- Add EMA (Exponential Moving Average).


klemen1999 (changes requested):
Would it make sense to have auxiliary segmentaion head part of the backbone (if `use_aux_head=True`) so we could then easily remove it from graph if needed (e.g. durgin export)? Because now user has to change use_aux_head=False in the config if they want to export optimally. CC: @kozlov721 also for thoughts

luxonis_train/nodes/backbones/ddrnet.py
```diff
@@ -0,0 +1,896 @@
...
+        self.layer4 = _make_layer(
+            block=block,
+            in_planes=width * 4,
+            planes=width * 8,
+            num_blocks=layers[3],
+            stride=2,
+        )
+
+
+class DDRNet(BaseNode[Tensor, list[Tensor]]):
```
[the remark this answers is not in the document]
kozlov721: I'd suggest making a submodule for the `ddrnet` with `ddrnet/blocks.py` containing just the blocks and `ddrnet/ddrnet.py` containing the main class. Check [nodes@cleanup](https://github.com/luxonis/luxonis-train/tree/cleanup/luxonis_train/nodes/backbones)

kozlov721:
> Would it make sense to have auxiliary segmentaion head part of the backbone (if `use_aux_head=True`) so we could then easily remove it from graph if needed (e.g. durgin export)? Because now user has to change use_aux_head=False in the config if they want to export optimally. CC: @kozlov721 also for thoughts

Yeah I think this would be a better solution. We could later add some more general support for train-only heads, but I think there's not gonna be many use-cases and they will mostly be solvable by making the head part of the backbone anyways.

Also, let's first merge #69, then sync this one and add a test case for the new predefined model to `tests.integration.test_simple.test_predefined_models`

luxonis_train/nodes/backbones/ddrnet/ddrnet.py
```diff
@@ -0,0 +1,298 @@
...
+        block: Type[nn.Module] = BasicResNetBlock,
+        skip_block: Type[nn.Module] = BasicResNetBlock,
+        layer5_block: Type[nn.Module] = Bottleneck,
+        layer5_bottleneck_expansion: int = 2,
+        spp_kernel_sizes: list[int] = None,
+        spp_strides: list[int] = None,
+        layer3_repeats: int = 1,
+        planes: int = 32,
+        layers: list[int] = None,
+        input_channels: int = 3,
```
kozlov721, line 43: There's already a property in `BaseNode` called `in_channels` that gives you the right number of input channels. It doesn't (shouldn't) have to be specified by the user in parameters.

luxonis_train/nodes/heads/ddrnet_segmentation_head.py
```diff
@@ -0,0 +1,94 @@
...
+
+from luxonis_train.nodes.base_node import BaseNode
+from luxonis_train.nodes.blocks import ConvModule
+from luxonis_train.utils.types import LabelType
+
+
+class DDRNetSegmentationHead(BaseNode[Tensor, Tensor]):
+    in_height: int
+    in_channels: int
+    tasks: list[LabelType] = [LabelType.SEGMENTATION]
```
kozlov721, line 20: `attach_index` should be defined here and be constant (the `BaseModel` takes care of changing it if provided by the user)

luxonis_train/nodes/heads/ddrnet_segmentation_head.py
```diff
@@ -0,0 +1,94 @@
...
+        @type inter_mode: str
+        @param inter_mode: Upsampling method. One of nearest, linear, bilinear, bicubic,
+            trilinear, area or pixel_shuffle. If pixel_shuffle is set, nn.PixelShuffle
+            is used for scaling. Defaults to "bilinear".
+        @type attach_index: int
+        @param attach_index: Index at which to attach. Defaults to 0.
+        """
+        self.attach_index = attach_index
+        super().__init__(**kwargs)
+        self.scale_factor = scale_factor
```
kozlov721, line 52: Not sure if it's the same thing, but there's a helper function `luxonis_train.utils.general.infer_upscale_factor` that will compute the upscale factor based on the input shape and original image shape.
Typical usage would be:
```python
h, w = self.original_in_shape[1:]
upsacle_factor = infer_upscale_factor((self.in_height, self.in_width), (h, w))
```
where `original_in_shape`, `in_height`, and `in_widht` are again already defined `BaseNode` properties.

luxonis_train/nodes/heads/ddrnet_segmentation_head.py
```diff
@@ -0,0 +1,94 @@
...
+
+        if inter_mode == "pixel_shuffle":
+            assert (
```
kozlov721, line 55: `ValueError` instead of `assert`

luxonis_train/nodes/heads/ddrnet_segmentation_head.py
```diff
@@ -0,0 +1,94 @@
...
+                inter_planes,
+                num_classes,
+                kernel_size=1,
+                padding=0,
+                bias=True,
+                activation=nn.Identity(),
+            )
+            self.upscale = nn.Upsample(scale_factor=scale_factor, mode=inter_mode)
+
+    def forward(self, x: Tensor) -> Tensor:
```
kozlov721, line 89: Nitpick: the argument should be named `inputs` so it's consistent with the other modules.

luxonis_train/nodes/heads/ddrnet_segmentation_head.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+class DDRNetSegmentationHead(BaseNode[Tensor, Tensor]):
+    in_height: int
+    in_channels: int
+    tasks: list[LabelType] = [LabelType.SEGMENTATION]
+
+    def __init__(
+        self,
+        num_classes: int,
```
kozlov721, line 24: Similarly to above, there are already defined properties `n_classes` and `in_channels` that can be used instead of `num_classes` and `in_planes` (and I suppose the default value of `inter_planes` should be then computed based on `in_channels`).

luxonis_train/nodes/backbones/ddrnet/ddrnet.py
```diff
@@ -0,0 +1,298 @@
+"""DDRNet backbone.
```
kozlov721, line 1: This can be moved to the class docstring. Also add the original source and paper links under `@see` tags .
(like `@see: U{Paper <https://...>}`)

luxonis_train/nodes/backbones/ddrnet/blocks.py
```diff
@@ -0,0 +1,574 @@
...
+from torch.nn import functional as F
+
+from luxonis_train.nodes.blocks import ConvModule
+
+
+class DropPath(nn.Module):
+    """Drop paths (Stochastic Depth) per sample, when applied in the main path of
+    residual blocks.
+
+    Intended usage of this block is as follows:
```
kozlov721, line 21: There should be two colons and the code block should be indented.

luxonis_train/nodes/backbones/ddrnet/blocks.py
```diff
@@ -0,0 +1,574 @@
...
+
+    >>> class ResNetBlock(nn.Module):
+    >>>   def __init__(self, ..., drop_path_rate: float):
+    >>>     self.drop_path = DropPath(drop_path_rate)
+    >>>
+    >>>   def forward(self, x):
+    >>>     return x + self.drop_path(self.conv_bn_act(x))
+
+    Code taken from TIMM (https://github.com/rwightman/pytorch-image-models), Apache License 2.0.
```
kozlov721, line 30:
```
@see: U{Original code <url>}
@license: U{Apache License, Version 2.0 <url to license in source repo>}
```

luxonis_train/nodes/backbones/ddrnet/blocks.py
```diff
@@ -0,0 +1,574 @@
...
+    def forward(self, x: Tensor) -> Tensor:
+        if self.drop_prob == 0.0 or not self.training:
+            return x
+        return drop_path(x, self.drop_prob, self.scale_by_keep)
+
+    def extra_repr(self) -> str:
+        return f"drop_prob={round(self.drop_prob, 3):0.3f}"
+
+
+class BasicResNetBlock(nn.Module):
```
kozlov721, line 57: This should probably go under `luxonis_ml.nodes.blocks` instead of here.

luxonis_train/nodes/backbones/ddrnet/blocks.py
```diff
@@ -0,0 +1,574 @@
...
+        out = F.relu(self.bn1(self.conv1(x)))
+        out = self.bn2(self.conv2(out))
+        out = self.drop_path(out)
+        out += self.shortcut(x)
+        if self.final_relu:
+            out = F.relu(out)
+        return out
+
+
+class Bottleneck(nn.Module):
```
kozlov721, line 120: The same as for resnet block

luxonis_train/nodes/backbones/ddrnet/blocks.py
```diff
@@ -0,0 +1,574 @@
...
+        x_list = [self.branches[0](x)]
+
+        for i in range(1, len(self.branches)):
+            x_list.append(self.branches[i]([x, x_list[i - 1]]))
+
+        out = self.compression(torch.cat(x_list, dim=1)) + self.shortcut(x)
+        return out
+
+
+class UpscaleOnline(nn.Module):
```
kozlov721, line 343: Same as resnet block

luxonis_train/nodes/heads/ddrnet_segmentation_head.py
```diff
@@ -0,0 +1,94 @@
+"""DDRNet segmentation head.
```
kozlov721, line 1: Same as module docstring in `drrnet.py`

kozlov721 (changes requested):
Looks good in general. Just left a few comments on docs and on how to better use the `BaseNode` API

luxonis_train/nodes/heads/ddrnet_segmentation_head.py
sokovninn: @kozlov721 for this model `scale_factor` is used to scale the output to the input shape, not to the original image shape.

luxonis_train/nodes/heads/ddrnet_segmentation_head.py
sokovninn: I would keep `in_planes`, `inter_planes` and `scale_factor` as separate arguments, as in the original implementation. Not sure how to use `in_channels` instead of `in_planes`.

luxonis_train/nodes/heads/ddrnet_segmentation_head.py
kozlov721: Yeah that's what is meant by the original in shape. It means the input shape of the entire model. I can see how the name might be a bit confusing, implying it's the original shape of the raw image data before augmentations etc.

(just a little correction for the above snippet, the actual upscale factor would be `2 ** infer_upsalce_factor()`).

luxonis_train/nodes/heads/ddrnet_segmentation_head.py
kozlov721: `BaseNode.in_channels` is exactly the same thing as what you're using the `in_planes` for, that is the number of channels of the input to the head. Only there is the additional value of being always set to the right number (in order for the shapes in the convolution layers to match) instead of relying on the user to provide it (who in our case is neither a programmer nor knowledgeable in ML). 

So, in short you just replace `in_planes` with `self.in_channels` and it will work just the same but won't break if the shape of the input changes. And you can also remove the `in_planes` computation from the predefined model. 

Specific example for why to do it this way; if I were to use this head with the resnet backbone, it will fail because the backbone produces outputs with 2048 channels, but the head expects 128. So I as a user would have to figure out what to change and how in order for the model to work. However, by using the `BaseNode` properties, I can attach this head to any backbone and it simply :magic_wand: _works_ :magic_wand: 

As a rule of thumb; any purely technical parameter that must be set correctly in order for the shapes to match (like `in_planes` or `scale_factor`) should not be required from the user.

configs/ddrnet_segmentation_model.yaml
```diff
@@ -0,0 +1,43 @@
+# DDRNet-23-slim model for segmentation
+# Refer to here for optimal hyperparameters for this model: https://github.com/Deci-AI/super-gradients/blob/4797c974c7c445d12e2575c468848d9c3e04becd/src/super_gradients/recipes/cityscapes_ddrnet.yaml#L4
+
+model:
+  name: ddrnet_segmentation
+  predefined_model:
+    name: DDRNetSegmentationModel
+    params:
+      n_classes: 80
```
kozlov721, line 9: `n_classes` should not be set, they're inferred from the dataset

luxonis_train/models/predefined_models/ddrnet_segmentation_model.py
```diff
@@ -0,0 +1,91 @@
...
+)
+from luxonis_train.utils.types import Kwargs
+
+from .segmentation_model import SegmentationModel
+
+
+@dataclass
+class DDRNetSegmentationModel(SegmentationModel):
+    backbone: str = "DDRNet"
+    n_classes: int = 1
```
kozlov721, line 15: There should be no default value for `n_classes`, only type annotation

kozlov721, line 15: `n_classes` should not be configurable

luxonis_train/models/predefined_models/ddrnet_segmentation_model.py
```diff
@@ -0,0 +1,91 @@
...
+    highres_planes: int = 64
```
kozlov721, line 16: This can be omitted (implicitly moved to `backbone_params`)

luxonis_train/models/predefined_models/ddrnet_segmentation_model.py
```diff
@@ -0,0 +1,91 @@
...
+    aux_head_params: Kwargs = field(default_factory=dict)
+
+    @property
+    def nodes(self) -> list[ModelNodeConfig]:
+        self.backbone_params.update({"highres_planes": self.highres_planes})
+        self.backbone_params.update(
+            {"layer5_bottleneck_expansion": self.layer5_bottleneck_expansion}
+        )
+
+        self.head_params.update(
```
kozlov721, line 27: When using the suggested `BaseNode` properties, the computation of `in_planes` is no longer necessary

luxonis_train/nodes/backbones/ddrnet/blocks.py
```diff
@@ -0,0 +1,350 @@
...
+        if stride != 1:
+            self.process = nn.Sequential(
+                nn.BatchNorm2d(branch_planes),
+                nn.ReLU(inplace=True),
+                nn.Conv2d(
+                    branch_planes, branch_planes, kernel_size=3, padding=1, bias=False
+                ),
+            )
+
+    def forward(self, x: Tensor) -> Tensor:
```
kozlov721, line 68: The type should be `Tensor | list[Tensor]`

luxonis_train/nodes/backbones/ddrnet/blocks.py
```diff
@@ -0,0 +1,350 @@
...
+
+from luxonis_train.nodes.blocks import ConvModule, UpscaleOnline
+
+
+class DAPPMBranch(nn.Module):
+    def __init__(
+        self,
+        kernel_size: int,
+        stride: int,
+        in_planes: int,
```
kozlov721, line 21: More general note, but we should rename `planes` to channels everywhere so it's consistent with the rest of the codebase.

luxonis_train/nodes/backbones/ddrnet/ddrnet.py
```diff
@@ -0,0 +1,295 @@
...
+            kernel_sizes=spp_kernel_sizes,
+            strides=spp_strides,
+        )
+
+        self.highres_planes = highres_planes
+        self.layer5_bottleneck_expansion = layer5_bottleneck_expansion
+        self.init_params()
+
+    @property
+    def backbone(self):
```
kozlov721, line 226: This looks like a dead code

luxonis_train/nodes/backbones/ddrnet/ddrnet.py
```diff
@@ -0,0 +1,295 @@
...
+        out_layer5_skip = self.layer5_skip(self.relu(x_skip))
+
+        x = self.upscale(
+            self.spp(self.layer5(self.relu(x))), height_output, width_output
+        )
+
+        x = x + out_layer5_skip
+
+        if self._use_aux_heads:
+            return [x, x_extra]
```
kozlov721, line 283: The order would be better reversed (`[x_extra, x]`) for a better compatibility with different heads. See the comment under `attach_index` for `DDRNetSegmentationHead`

luxonis_train/nodes/heads/ddrnet_segmentation_head.py
```diff
@@ -0,0 +1,106 @@
...
+from luxonis_train.utils.types import LabelType
+
+logger = logging.getLogger(__name__)
+
+
+class DDRNetSegmentationHead(BaseNode[Tensor, Tensor]):
+    in_height: int
+    n_classes: int
+    tasks: list[LabelType] = [LabelType.SEGMENTATION]
+    attach_index: int = 0
```
kozlov721, line 17: It's safer for the `attach_index` to be negative for better compatibility with other backbones. The normal head would be attached to `-1` and the auxiliary to `-2`.

luxonis_train/nodes/heads/ddrnet_segmentation_head.py
```diff
@@ -0,0 +1,106 @@
...
+
+        @warning: The replacement is destructive and cannot be undone.
+        @type mode: bool
+        @param mode: Whether to set the export mode to True or False. Defaults to True.
+        """
+        super().set_export_mode(mode)
+        if self.export and self.attach_index != 0:
+            logger.info("Removing the auxiliary head.")
+
+            self.forward = lambda x: x
```
kozlov721, line 106: Not sure how much it would increase the throughput of the model, but we could use something even simpler like `self.forward = lambda x: torch.tensor(0)` instead.

luxonis_train/nodes/heads/ddrnet_segmentation_head.py
sokovninn: Got, now it makes sense for me:)

kozlov721 (approved):
LGTM

klemen1999 (approved):
LGTM

PR #76, opened 2024-09-20
Upsample mode parameter for UpBlock

sokovninn:
This PR adds an `upsample_mode` parameter to the `UpBlock`, which is used in the default segmentation head, to avoid checkerboard artifacts caused by the `nn.ConvTranspose2d `layer.

Now, `nn.Upsample `is the default choice in the `UpBlock`.

### DIFF
--- a/luxonis_train/nodes/blocks/blocks.py
+++ b/luxonis_train/nodes/blocks/blocks.py
@@ -1,5 +1,5 @@
 import math
-from typing import TypeVar
+from typing import Literal, TypeVar
 
 import numpy as np
 import torch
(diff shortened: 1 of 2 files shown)


luxonis_train/nodes/blocks/blocks.py
```diff
@@ -149,18 +153,48 @@ def __init__(
...
-            nn.ConvTranspose2d(
-                in_channels,
-                out_channels,
-                kernel_size=kernel_size,
-                stride=stride,
-            ),
-            ConvModule(out_channels, out_channels, kernel_size=3, padding=1),
+        layers = []
+
+        if upsample_mode == "transpose":
```
kozlov721, line 169: Should this be `"cov_transpose"`?

sokovninn: Oh, right, thanks!

kozlov721 (approved):
LGTM

klemen1999 (approved):

PR #79, opened 2024-09-21
Simplified Node Configuration

kozlov721:
- `metrics`, `losses`, and `visualizers` can be defined under their respective nodes instead of in separate fields under `model`
  - No need to specify the `attached_to` field for such modules
  - The old way is still supported, but raises a deprecation warning
- `node.inputs` not required for simple models following `head -> (neck?) -> head1, head2, ...` topology
  - The nodes must be defined in the config in this exact order 
  - For multihead networks, the heads must have `"Head"` in their name for this to work

### Old Config

```yaml
model:
  name: resnet50_classification
  nodes:
    - name: ResNet
      params:
        variant: "50"
        download_weights: True

    - name: ClassificationHead
      inputs:
        - ResNet

  losses:
    - name: CrossEntropyLoss
      attached_to: ClassificationHead

  metrics:
    - name: Accuracy
      is_main_metric: true
      attached_to: ClassificationHead
    - name: Recall
      attached_to: ClassificationHead


  visualizers:
    - name: ClassificationVisualizer
      attached_to: ClassificationHead
```

### New Config

```yaml
model:
  name: resnet50_classification
  nodes:
    - name: ResNet
      params:
        variant: "50"
        download_weights: True

    - name: ClassificationHead

      losses:
        name: CrossEntropyLoss

      metrics:
        - name: Accuracy
          is_main_metric: true
        - name: Recall

      visualizers:
        name: ClassificationVisualizer
```


luxonis_train/config/config.py
```diff
@@ -82,6 +83,44 @@ class ModelConfig(BaseModelExtraForbid):
...
+            return nodes
+        names = []
+        last_body_index: int | None = None
+        for i, node in enumerate(nodes):
+            name = node.get("alias", node.get("name"))
+            if name is None:
+                raise ValueError(
+                    f"Node {i} does not specify the `name` field."
+                )
+            if "Head" in name and last_body_index is None:
```
klemen1999, line 100: This requires "Head" to be part of the class name to automatically work right?

klemen1999 (approved):
LGTM

luxonis_train/config/config.py
kozlov721: Yeah, more specifically when first head is encountered, it assumes the node before was a neck and all subsequent nodes will be connected to it. By default, it just connects it to the previous one. So the `"Head"` in name is only required if you have multiple heads. For simply linear networks it will work no matter the names.

PR #82, opened 2024-09-26
Feat/inference

JSabadin:
## Inference Functionality Updates

- Refactored the `infer()` method with the following helper methods:
  - `_process_single_image`: Handles inference for a single image file.
  - `_process_directory_images`: Handles inference for image files in a directory.
  - `_process_dataset_images`: Handles inference on dataset images.
  - `_prepare_labels`: Prepares labels for various tasks.

- The `infer()` method now supports:
  - **Single Image File** inference.
  - **Directory of Images** inference.
  - **Luxonis Dataset** inference

### DIFF
--- a/luxonis_train/__main__.py
+++ b/luxonis_train/__main__.py
@@ -2,7 +2,7 @@ import tempfile
 from enum import Enum
 from importlib.metadata import version
 from pathlib import Path
-from typing import Annotated, Optional, Union
+from typing import Annotated
 
 import typer
 import yaml
@@ -25,7 +25,7 @@ app = typer.Typer(
 
 
 ConfigType = Annotated[
-    Optional[str],
+    str | None,
     typer.Option(
         help="Path to the configuration file.",
         show_default=False,
@@ -34,7 +34,7 @@ ConfigType = Annotated[
 ]
 
 OptsType = Annotated[
-    Optional[list[str]],
+    list[str] | None,
     typer.Argument(
         help="A list of optional CLI overrides of the config file.",
         show_default=False,
@@ -46,13 +46,15 @@ ViewType = Annotated[
 ]
 
 SaveDirType = Annotated[
-    Optional[Path],
+    Path | None,
     typer.Option(help="Where to save the inference results."),
 ]
 
 ImgPathType = Annotated[
-    Optional[str],
-    typer.Option(help="Path to an image file or a directory containing images for inference.")
+    str | None,
+    typer.Option(
+        help="Path to an image file or a directory containing images for inference."
+    ),
 ]
 
 
@@ -60,7 +62,7 @@ ImgPathType = Annotated[
 def train(
     config: ConfigType = None,
     resume: Annotated[
-        Optional[str],
+        str | None,
         typer.Option(help="Resume training from this checkpoint."),
     ] = None,
     opts: OptsType = None,
@@ -110,7 +112,9 @@ def infer(
     """Run inference."""
     from luxonis_train.core import LuxonisModel
 
-    LuxonisModel(config, opts).infer(view=view.value, save_dir=save_dir, img_path=img_path)
+    LuxonisModel(config, opts).infer(
+        view=view.value, save_dir=save_dir, img_path=img_path
+    )
 
 
 @app.command()
@@ -206,7 +210,7 @@ def common(
         ),
     ] = False,
     source: Annotated[
-        Optional[Path],
+        Path | None,
         typer.Option(
             help="Path to a python file with custom components. "
             "Will be sourced before running the command.",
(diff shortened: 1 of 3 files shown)


luxonis_train/core/core.py
```diff
@@ -17,7 +18,7 @@
 from luxonis_ml.nn_archive.config import CONFIG_VERSION
 from luxonis_ml.utils import LuxonisFileSystem, reset_logging, setup_logging
 from typeguard import typechecked
-
+from luxonis_ml.data import LabelType
```
kozlov721, line 21: Use `luxonis_train.enums.TaskType` instead. (see #78)

luxonis_train/core/core.py
```diff
@@ -419,6 +420,7 @@ def infer(
         self,
         view: Literal["train", "val", "test"] = "val",
         save_dir: str | Path | None = None,
+        img_path: Optional[str] = None
```
kozlov721, line 423: Instead of `Optional[str]`, you should use `str | None` (`Optional` is deprecated in python 3.10)

luxonis_train/core/core.py
```diff
@@ -429,16 +431,68 @@ def infer(
...
+        if img_path:
+            img_path_obj = Path(img_path)
+            if img_path_obj.is_file():
+                self._process_single_image(img_path_obj, view, save_dir)
+            elif img_path_obj.is_dir():
+                self._process_directory_images(img_path_obj, view, save_dir)
+        else:
+            self._process_dataset_images(view, save_dir)
+
+    def _process_single_image(self, img_path: Path, view: str, save_dir: Optional[str | Path]) -> None:
```
kozlov721, line 449: It might be better to move these methods to `luxonis_train.core.utils.infer_utils`

luxonis_train/__main__.py
```diff
@@ -50,6 +50,11 @@ class _ViewType(str, Enum):
     typer.Option(help="Where to save the inference results."),
 ]
 
+ImgPathType = Annotated[
+    Optional[str],
```
kozlov721, line 54: `str | None`

luxonis_train/core/core.py
```diff
@@ -429,16 +431,68 @@ def infer(
...
+    def _process_dataset_images(self, view: str, save_dir: Optional[str | Path]) -> None:
+        """Handles the inference on dataset images."""
         for inputs, labels in self.pytorch_loaders[view]:
             images = get_unnormalized_images(self.cfg, inputs)
             outputs = self.lightning_module.forward(
                 inputs, labels, images=images, compute_visualizations=True
             )
             render_visualizations(outputs.visualizations, save_dir)
 
+    def _prepare_labels(self, view: str, img_shape: tuple) -> tuple:
```
kozlov721, line 477: `_create_dummy_labels` might be a better name

kozlov721 (approved):
Left a few nit-picky comments, otherwise LGTM!

luxonis_train/core/utils/infer_utils.py
```diff
@@ -28,3 +33,70 @@ def render_visualizations(
...
         if cv2.waitKey(0) == ord("q"):
             exit()
+
+
+def process_single_image(
+    model, img_path: Path, view: str, save_dir: str | Path | None
+) -> None:
+    """Handles the inference on a single image."""
+    img = cv2.cvtColor(cv2.imread(str(img_path)), cv2.COLOR_BGR2RGB)
+    img, _ = model.val_augmentations([(img, {})])
```
klemen1999, line 43: Since we already have `view` passed here can we also set augmentations depending on this value (if train use train augs otherwise val)? Could be nice for cases where you maybe want to infer on images with some specific augmentations.

luxonis_train/core/core.py
```diff
@@ -429,15 +433,20 @@ def infer(
...
-        for inputs, labels in self.pytorch_loaders[view]:
-            images = get_unnormalized_images(self.cfg, inputs)
-            outputs = self.lightning_module.forward(
-                inputs, labels, images=images, compute_visualizations=True
-            )
-            render_visualizations(outputs.visualizations, save_dir)
+        if img_path:
+            img_path_obj = Path(img_path)
+            if img_path_obj.is_file():
+                process_single_image(self, img_path_obj, view, save_dir)
```
klemen1999, line 445: Can these two be merged: single image can be packaged in a list of len=1 and represetn a "directory of images". And then dummy labels can be generated only once for whole directory and process_single_image would take them as parameter. This would optimize the label generation a bit as currently we re-generate them for every image in the directory. Thoughts?

klemen1999 (approved):
LGTM, left some comments

klemen1999 (approved):
LGTM

PR #85, opened 2024-10-01
Lock Down Sample Configs

JSabadin:
# Configuration Updates

### Summary
This PR introduces the creation of **light** and **heavy** configuration files for various tasks, including segmentation, detection, keypoints, and classification.

### Key Changes
- Added light and heavy model configurations for:
  - Segmentation
  - Detection
  - Keypoints
  - Classification


configs/coco_model.yaml
```diff
@@ -105,7 +96,7 @@ trainer:
   n_sanity_val_steps: 1
   profiler: null
   verbose: True
-  batch_size: 4
+  batch_size: 8
   accumulate_grad_batches: 1
   epochs: &epochs 200
   n_workers: 8
```
klemen1999, line 102: Can we double check if `train_metrics_interval` parameter is even used anywhere? Because I believe we don't actually call a function where this is used. If that is the case then we should remove this config parameter (and the function)

luxonis_train/config/predefined_models/classification_model.py
```diff
@@ -14,19 +14,32 @@
 
 @dataclass
 class ClassificationModel(BasePredefinedModel):
-    backbone: str = "MicroNet"
+    variant: str | None = None
```
klemen1999, line 17: Can we make the variants more similar to how we handle them in e.g. EfficientRep? So make predefined model an actual class (instead of dataclass) with all the parameters being None by default (except variant="lite" by default). Then have VariantLiteral type ["lite", "heavy"] and a function that gets specific parameters based on variant and uses them if others aren't specifid (e.g. uses "ResNet-18" as backbone if variant="lite" and no backbone is set, but if backbone is explicitly set then use that one.

The same can be done for all ther predefined models.

luxonis_train/config/predefined_models/segmentation_model.py
```diff
@@ -14,36 +14,61 @@
 
 @dataclass
 class SegmentationModel(BasePredefinedModel):
-    backbone: str = "MicroNet"
+    variant: str | None = None
```
klemen1999, line 17: Keep backbone as parameter (same as with ClassificationModel) so it can be explicitly set.

klemen1999 (commented):
Left some comments. Overall let's try to handle variants similarly as we do with e.g. EfficientRep or RepPANNeck.

luxonis_train/attached_modules/losses/efficient_keypoint_bbox_loss.py
```diff
@@ -33,10 +33,10 @@ def __init__(
         n_warmup_epochs: int = 4,
         iou_type: IoUType = "giou",
         reduction: Literal["sum", "mean"] = "mean",
-        class_loss_weight: float = 1.0,
-        iou_loss_weight: float = 2.5,
+        class_loss_weight: float = 0.5,
```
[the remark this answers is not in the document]
JSabadin: Those are based on YOLOv8.

configs/classification_heavy_model.yaml
```diff
@@ -1,24 +1,24 @@
-# Example configuration for training a predefined detection model
+# Example configuration for training a predefined heavy classification model
 
 model:
-  name: coco_detection
+  name: segmentation_light
```
kozlov721, line 4: Should this be `classification_heavy`?

configs/classification_light_model.yaml
```diff
@@ -1,29 +1,24 @@
-# Example configuration for training a predefined segmentation model
+# Example configuration for training a predefined light classification model
 
 model:
-  name: cifar10_classification
+  name: segmentation_light
```
kozlov721, line 4: `classification_light`

configs/example_export.yaml
```diff
@@ -4,23 +4,22 @@ model:
   name: coco_segmentation
```
kozlov721, line 4: `detection_export`?

configs/classification_heavy_model.yaml
```diff
@@ -1,24 +1,24 @@
...
 loader:
   params:
-    dataset_name: coco_test
+    dataset_name: cifar10_test
 
 trainer:
   preprocessing:
-    train_image_size: [&height 256, &width 320]
-    keep_aspect_ratio: False
+    train_image_size: [&height 384, &width 512]
```
kozlov721, line 16: We don't need to capture `height`, `width`, and `epochs` in all the yaml files. We can do it only when the captures are actually used (we're using `epochs` for `T_max` in some schedulers and `width` and `height` for mosaic augmentation)

luxonis_train/config/predefined_models/classification_model.py
```diff
@@ -11,19 +12,63 @@
 
 from .base_predefined_model import BasePredefinedModel
 
+VariantLiteral: TypeAlias = Literal["lite", "heavy"]
```
kozlov721, line 15: This should be `"light"`

kozlov721 (commented):
Left some comments. Additionally, the `test_predefined_models` in `tests/integration/test_simple.py` will need to be changed to use the new configs.
Otherwise LGTM

klemen1999:
We should also update the README under `predefined_models` accordingly

PR #86, opened 2024-10-01
[Feat] Added name attribute to HeadType

klemen1999:
- Added `name` attribute to head when automatically generating NNArchive (Related to [this luxonis-ml PR](https://github.com/luxonis/luxonis-ml/pull/176)) 
- Fixed output names for EfficientBboxHead to fit DAI requirements. This is only triggered if user doesn't specify the output names through config.


kozlov721 (approved):
LGTM

PR #87, opened 2024-10-02
Updates to the READMEs

klemen1999:
Updated READMEs with new nodes and changes to the config parameters. 

I also removed mention of `ImplicitKeypointBBoxHead` and `ImplicitKeypointBBoxLoss` in the READMEs since we don't want to promote to use of this head. Users should rather use `EfficientKeypointBBoxHead` because it is faster and more stable for training.


kozlov721 (approved):
LGTM

PR #100, opened 2024-10-08
Inference Fix

kozlov721:
- Updated visualizers so they don't require labels
- Fixed rendering of predictions for `batch_size` larger than 1
- Added tests for all inference options
- Fixed failing `pre-commit`


klemen1999 (approved):
LGTM

PR #111, opened 2024-10-10
SequentialLR Scheduler

JSabadin:
# SequentialLR Support for Training

## Overview

This PR adds support for the `SequentialLR` learning rate scheduler in the training configuration. The `SequentialLR` scheduler allows for chaining multiple learning rate schedulers, making it possible to implement a warm-up phase followed by a more complex learning rate schedule during training.

## Changes

- **SequentialLR Integration**: You can now use the `SequentialLR` scheduler in the training configuration to combine multiple schedulers, such as a warm-up phase (`LinearLR`) followed by `CosineAnnealingLR`.
- **Configuration Example**:
  ```yaml
  scheduler:
    name: SequentialLR
    params:
      schedulers:
        - name: LinearLR               # Warmup phase
          params:
            start_factor: 0.1
            total_iters: 10
        - name: CosineAnnealingLR      # After warmup, switch to Cosine Annealing
          params:
            T_max: 299
            eta_min: 0.00001
      milestones: [10]  # Switch to CosineAnnealingLR after 10 iterations

### DIFF
--- a/luxonis_train/models/luxonis_lightning.py
+++ b/luxonis_train/models/luxonis_lightning.py
@@ -813,8 +813,32 @@ class LuxonisLightningModule(pl.LightningModule):
         }
         optimizer = OPTIMIZERS.get(cfg_optimizer.name)(**optim_params)
 
-        scheduler_params = cfg_scheduler.params | {"optimizer": optimizer}
-        scheduler = SCHEDULERS.get(cfg_scheduler.name)(**scheduler_params)
+        def get_scheduler(scheduler_cfg, optimizer):
+            scheduler_class = SCHEDULERS.get(
+                scheduler_cfg["name"]
+            )  # For dictionary access
+            scheduler_params = scheduler_cfg["params"] | {
+                "optimizer": optimizer
+            }  # Dictionary access for params
+            return scheduler_class(**scheduler_params)
+
+        if cfg_scheduler.name == "SequentialLR":
+            schedulers_list = [
+                get_scheduler(scheduler_cfg, optimizer)
+                for scheduler_cfg in cfg_scheduler.params["schedulers"]
+            ]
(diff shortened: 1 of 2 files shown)


kozlov721 (approved):
LGTM

klemen1999 (commented):
LGTM, let's add some tests for this - I suppose we need to test `configure_optimizers` function as a whole by trying different optimizer and scheduler configurations. 
CC: @kozlov721 on where would be the best way to put them. IMO under unittests but for them to run we need a LuxonisLighnitngModule instance I think so not sure if this then falls under integration tests?

kozlov721:
> LGTM, let's add some tests for this - I suppose we need to test `configure_optimizers` function as a whole by trying different optimizer and scheduler configurations. CC: @kozlov721 on where would be the best way to put them. IMO under unittests but for them to run we need a LuxonisLighnitngModule instance I think so not sure if this then falls under integration tests?

I don't think this can be easily unit-tested. We can add `test_scheduler` to integration tests that would just start a training with a simple config and 2 epochs with a change of scheduler after the first epoch.

klemen1999:
> We can add test_scheduler to integration tests that would just start a training with a simple config and 2 epochs with a change of scheduler after the first epoch.

Yes this makes sense. But can we use like a 10 image dataset (as small as possible) for such traning runs so we don't increase the test times? We should use a very small dataset for all tests where we don't actually care about the performance and we are just testing funcionality.

JSabadin:
Will create a test for this as mentioned above. We can also reduce the CIFAR dataset to just about 10 images [here](https://github.com/luxonis/luxonis-train/blob/6146da6422388ed97945d42bcd1c4625c2bc9d68/tests/integration/conftest.py#L232).

PR #115, opened 2024-10-17
Add gradient visualization during validation

sokovninn:
This pull request introduces a new callback for visualizing gradients using Grad-CAM during validation in the `luxonis_train` package. The most important changes include the addition of the `GradCamCallback` class, updates to the callback registration, and documentation updates.

### New Callback Implementation:

* [`luxonis_train/callbacks/gradcam_visializer.py`](diffhunk://#diff-e08ea37a71509571a354203074563c0ad03ab687865bdf8686f95fd66655eddaR1-R205): Added the `GradCamCallback` class and its supporting `ModelWrapper` class to enable gradient visualization using Grad-CAM during validation. 

### Callback Registration:

* [`luxonis_train/callbacks/__init__.py`](diffhunk://#diff-9fb27879fda34ab85599e83a085fe46f99a42acbaeea39b48c714be901cce3ebR18): Registered the `GradCamCallback` class to make it available for use in the training pipeline. 

### Documentation:

* [`luxonis_train/callbacks/README.md`](diffhunk://#diff-51d568473917a921fe145cfef11d0827a6a3e277da0c8da439989792acd411f8R68-R80): Updated the documentation to include a description of the `GradCamCallback` and its parameters.

### DIFF
--- a/luxonis_train/callbacks/README.md
+++ b/luxonis_train/callbacks/README.md
@@ -65,3 +65,16 @@ Callback to perform a test run at the end of the training.
 ## `UploadCheckpoint`
 
 Callback that uploads currently the best checkpoint (based on validation loss) to the tracker location - where all other logs are stored.
+
+## `GradCamCallback`
+
+Callback to visualize gradients using Grad-CAM. Works only during validation.
+
+**Parameters:**
+
+| Key             | Type  | Default value      | Description                                          |
+| --------------- | ----- | ------------------ | ---------------------------------------------------- |
+| `target_layer`  | `int` | -                  | Layer to visualize gradients.                        |
+| `class_idx`     | `int` | 0                  | Index of the class for visualization. Defaults to 0. |
+| `log_n_batches` | `int` | 1                  | Number of batches to log. Defaults to 1.             |
+| `task`          | `str` | `"classification"` | The type of task. Defaults to "classification".      |
--- a/luxonis_train/callbacks/README.md
+++ b/luxonis_train/callbacks/README.md
@@ -66,7 +66,6 @@ Callback to perform a test run at the end of the training.
 
 Callback that uploads currently the best checkpoint (based on validation loss) to the tracker location - where all other logs are stored.
 
-
 ## `GradCamCallback`
 
 Callback to visualize gradients using Grad-CAM. Works only during validation.
--- a/luxonis_train/callbacks/gradcam_visializer.py
+++ b/luxonis_train/callbacks/gradcam_visializer.py
@@ -1,0 +1,123 @@
+import pytorch_lightning as pl
+import torch.nn.functional as F
+import torch
+from torchvision import transforms
+import matplotlib.pyplot as plt
+import numpy as np
+
+from pytorch_grad_cam import GradCAM
+from pytorch_grad_cam.utils.image import show_cam_on_image
+from pytorch_grad_cam.utils.model_targets import SemanticSegmentationTarget
+
+
+from luxonis_train.attached_modules.visualizers import (
+    combine_visualizations,
+    get_denormalized_images,
+)
+
+class SegmentationWrapper(pl.LightningModule):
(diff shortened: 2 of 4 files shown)


luxonis_train/callbacks/README.md
```diff
@@ -65,3 +65,16 @@ Callback to perform a test run at the end of the training.
...
+## `GradCamCallback`
+
+Callback to visualize gradients using Grad-CAM. Works only during validation.
+
+**Parameters:**
+
+| Key             | Type  | Default value      | Description                                          |
+| --------------- | ----- | ------------------ | ---------------------------------------------------- |
+| `target_layer`  | `int` | -                  | Layer to visualize gradients.                        |
+| `class_idx`     | `int` | 0                  | Index of the class for visualization. Defaults to 0. |
```
klemen1999, line 78: Should we make it we visualize for all classes by default (e.g. if -1) and then for specific one if set to some positive integer? Also we'll want to check if this number is > than number of all classes.

luxonis_train/callbacks/README.md
```diff
@@ -65,3 +65,16 @@ Callback to perform a test run at the end of the training.
...
+
+## `GradCamCallback`
+
+Callback to visualize gradients using Grad-CAM. Works only during validation.
+
+**Parameters:**
+
+| Key             | Type  | Default value      | Description                                          |
+| --------------- | ----- | ------------------ | ---------------------------------------------------- |
+| `target_layer`  | `int` | -                  | Layer to visualize gradients.                        |
```
klemen1999, line 77: Are there any guidelines we can put here as a note for users on which layers to choose? Also it might be hard to actually "guess" the index of the desired layer. Can we make this more user friendly?

requirements.txt
```diff
@@ -17,3 +17,4 @@ typer>=0.9.0
 mlflow>=2.10.0
 psutil>=5.0.0
 tabulate>=0.9.0
+grad-cam>=1.5.4
```
klemen1999, line 20: I wouldn't have always required. Do we want to put this requirement directly in the class? CC: also @kozlov721

luxonis_train/callbacks/gradcam_visializer.py
```diff
@@ -0,0 +1,205 @@
...
+        for zip_idx, (image, grayscale_cam) in enumerate(
+            zip(images, grayscale_cams)
+        ):
+            image = image / 255.0
+            image = image.transpose(1, 2, 0)
+            visualization = show_cam_on_image(
+                image, grayscale_cam, use_rgb=True
+            )
+
+            fig, ax = plt.subplots(1, 1)
```
klemen1999, line 197: Nitpick: Can we replace the use of matplotlib with cv2 so we don't need to create figures. There is `cv2.applyColorMap(im, cv2.COLORMAP_JET)` function which might suffice?

klemen1999 (commented):
Overall LGTM, left some comments

luxonis_train/callbacks/README.md
sokovninn: Initially, I had this option, but I decided to remove it because typically we want to examine/debug a specific class. If all classes are visualized, we can gain some insights, but it would be harder to understand where the issue lies. Moreover, adding more target classes will increase the callback processing time, as it will require more forward passes instead of a single batched pass due to the implementation of the Grad-CAM library. We can also add visualization of gradients separately for each or all target classes. However, I think we can keep advanced functionality for the future if it becomes necessary. And that’s a good point about the check.

requirements.txt
sokovninn: Yeah, moving it to the class makes sense

kozlov721: IMHO we don't really need to care about optimizing the dependencies of `luxonis-train` because it will be huge anyway

kozlov721 (approved):
LGTM

PR #116, opened 2024-10-18
Unsupervised Anomaly Detection

JSabadin:
### Overview
This PR introduces an unsupervised anomaly detection model that runs at 55 FPS (Heavy version)  on RVC4 and 10 FPS (Light version) RVC2 to improve performance. The model is one of the fastest available for anomaly detection.

The model is based on the work described in the following paper: [DRAEM - A Discriminatively Trained Reconstruction Embedding for Surface Anomaly Detection](https://openaccess.thecvf.com/content/ICCV2021/papers/Zavrtanik_DRAEM_-_A_Discriminatively_Trained_Reconstruction_Embedding_for_Surface_Anomaly_ICCV_2021_paper.pdf). 

### Implementation Details
One of the challenges during implementation was optimizing the Perlin noise generation, which is critical for the model's performance. This optimization allows the model to process the MTVEC dataset, which consists of 3,000 images, with only a 13-second overhead per epoch.
image

After manually implementing the noise in the forward pass of the backbone, we decided to move the process of generating noise and combining it with the image into a separate custom loader, eliminating the augmentation overhead.

### Config file example
To use this model, it must be initialized with the source directory of anomaly augmentations. This can be configured as shown below:

```yaml
model:
  name: DREAM
  nodes:
    - name: RecSubNet
    - name: DiscSubNetHead
      inputs: ["RecSubNet"]

  losses:
  - attached_to: DiscSubNetHead
    name: ReconstructionSegmentationLoss

  visualizers:
    - name: SegmentationVisualizer
      attached_to: DiscSubNetHead

loader:
  name: LuxonisLoaderPerlinNoise
  train_view: train
  val_view: train
  params:
    dataset_name: mvtec
    anomaly_source_path:  "C:\\path\\to\\dtd-r1.0.1\\dtd\\images\\"

trainer:
  preprocessing:
    train_image_size: [&height 256, &width 256]
    keep_aspect_ratio: False
    normalize:
      active: True

  batch_size: 4
  epochs: &epochs 700
  num_workers: 8
  validation_interval: 10
  num_log_images: 8

  optimizer:
    name: Adam
    params:
      lr: 0.0001
```
After 20 epochs (the official implementation uses 700 epochs), we obtained the results shown in the image.
image

### Performance
- **RVC4**: 55 FPS
- **RVC2**: 10 FPS

This model is a fast and efficient solution for anomaly detection.


luxonis_train/attached_modules/losses/reconstruction_segmentation_loss.py
```diff
@@ -0,0 +1,297 @@
...
+        ret = ssim_map.mean()
+    else:
+        ret = ssim_map.mean(1).mean(1).mean(1)
+
+    if full:
+        return ret, cs
+    return ret, ssim_map
+
+
+class FocalLoss(nn.Module):
```
klemen1999, line 209: Can this be replaced by the focal loss already in luxonis-train? Or can we extend the one already implemented to support functionality introduced in this one as well (e.g. label smoothing)?

luxonis_train/loaders/luxonis_perlin_loader_torch.py
```diff
@@ -0,0 +1,92 @@
...
+            )
+            if not self.anomaly_source_paths:
+                raise FileNotFoundError(
+                    "No .jpg files found at the specified path."
+                )
+        else:
+            raise ValueError("Invalid or unspecified anomaly source path.")
+        self.anomaly_source_path = anomaly_source_path
+        self.noise_prob = noise_prob
+        if mean is None:
```
klemen1999, line 53: we should have access to the Augmentation instance here. Can we get `Normalize` augmentation from there and apply it on the noise image as well instead of requiring mean and std params? And if Normalize isn't present then don't normalize.

luxonis_train/nodes/backbones/recsubnet/recsubnet.py
```diff
@@ -0,0 +1,32 @@
...
+    base_width: int
+
+    def __init__(
+        self,
+        in_channels=3,
+        out_channels=3,
+        base_width=128,
+        **kwargs,
+    ):
+        super().__init__(**kwargs)
```
klemen1999, line 22: Let's add docstrings and let's also include link to paper and original implementation.

luxonis_train/nodes/backbones/recsubnet/utils.py
```diff
@@ -0,0 +1,181 @@
+import glob
```
klemen1999, line 1: Seems like a copy of the perlin.py?

luxonis_train/nodes/heads/discsubnet_head/discsubnet_head.py
```diff
@@ -0,0 +1,67 @@
...
+    tasks: list[TaskType] = [TaskType.SEGMENTATION]
+
+    def __init__(
+        self,
+        in_channels: list[int] | int = 6,
+        out_channels: list[int] | int = 2,
+        base_channels: int = 64,
+        out_features: bool = False,
+        **kwargs,
+    ):
```
klemen1999, line 29: Same as with `RecSubNet` let's include docstrings and short explanation with credits

luxonis_train/loaders/luxonis_perlin_loader_torch.py
```diff
@@ -0,0 +1,92 @@
...
+                TaskType.SEGMENTATION,
+            )
+        else:
+            for task, (array, label_type) in labels.items():
+                tensor_labels[task] = (
+                    Tensor(array),
+                    TaskType(label_type.value),
+                )
+
+        return {self.image_source: tensor_img}, tensor_labels
```
klemen1999, line 92: Don't we need two image sources? One for image before perlin noise augmentation and one for after? Because if I understand correctly the output of the reconstructed sub network must always be compared to the image with no perlin noise augmentation. Where as in this case if perlin augmentation is performed then in the L_rec loss we compare output of the reconstructed image to image with perlin augmentation.

klemen1999 (commented):
Nice PR, left some comments. Also before we merge it we'll also want to add it to the docs, add tests and make a predefined a"nomaly detection model" with light and heavy version.

luxonis_train/loaders/luxonis_perlin_loader_torch.py
JSabadin: Yeah, you're absolutely right. I'll fix this!

luxonis_train/attached_modules/losses/softmax_focal_loss.py
```diff
@@ -34,24 +34,43 @@ def __init__(
...
 
-    def forward(self, predictions: Tensor, target: Tensor) -> Tensor:
-        ce_loss = self.ce_criterion.forward(predictions, target)
-        pt = torch.exp(-ce_loss)
-        loss = ce_loss * ((1 - pt) ** self.gamma) * self.alpha
+        if self.smooth is not None:
+            if self.smooth < 0 or self.smooth > 1.0:
+                raise ValueError("smooth value should be in [0,1]")
+
+    def forward(
```
klemen1999, line 53: Let's make sure that the output of this loss with smoothing turned off is same as with the previous implementation. And also let's make sure that it still works with e.g. segmentation model correctly.

We should keep naming of the parameters consistent across losses: `predictions`, `target`

luxonis_train/config/predefined_models/anomaly_detection_model.py
```diff
@@ -0,0 +1,123 @@
...
+        self.task_name = task_name or "anomaly_detection"
+
+    @property
+    def nodes(self) -> list[ModelNodeConfig]:
+        """Defines the model nodes, including RecSubNet and
+        DiscSubNetHead."""
+        return [
+            ModelNodeConfig(
+                name=self.backbone,
+                alias="rec_subnet",
```
klemen1999, line 76: Let's update this (and all other nodes, losses, metrics and visualizers) like we've done for other predefined models in #118

klemen1999 (approved):
Feel free to merge once comment is resolved

luxonis_train/nodes/backbones/recsubnet/blocks.py
```diff
@@ -0,0 +1,156 @@
+from torch import nn
```
kozlov721, line 1: Can you please add types to the blocks and remove `super` arguments?

luxonis_train/nodes/heads/discsubnet_head/blocks.py
```diff
@@ -0,0 +1,166 @@
...
+            kernel_size=3,
+            padding=1,
+            activation=nn.ReLU(inplace=True),
+        ),
+    )
+
+
+class Encoder(nn.Module):
+    def __init__(self, in_channels: int, base_width: int) -> None:
+        super(Encoder, self).__init__()
```
kozlov721, line 41: The `super` arguments are a relic from python 2, please remove them so we're consistent with the rest of the codebase

kozlov721 (approved):
LGTM, left some comments. Feel free to merge once the conversations are resolved.

PR #119, opened 2024-10-30
fix: call tracker.close()

JSabadin:
# Call `close()` on Tracker After Training Completion

## Summary
This update ensures `close()` is executed on the tracker after training finishes, aligning with recent changes in `luxonis-ml` ([This PR](https://github.com/luxonis/luxonis-ml/pull/193))


kozlov721 (approved):
LGTM, but we need to make a new release of `luxonis-ml` before we merge

PR #121, opened 2024-11-05
Run inference on the same device as used in the Trainer

sokovninn:
This PR includes:

- Added inference logic to the `LuxonisLightningModule`
- Dataset and image folder inference now run on the same device as used in the `Trainer`
- Loss is no longer computed during inference

The implementation of video inference has not changed.


kozlov721 (approved):
LGTM

klemen1999 (approved):
LGTM

PR #122, opened 2024-11-06
Updated how we access config version from luxonis-ml

klemen1999:
Since `CONFIG_VERSION` is now a string we need to update how we read it in luxonis-train


klemen1999:
**NOTE:** [This PR](https://github.com/luxonis/luxonis-ml/pull/200) needs to be merged in before

kozlov721 (approved):
LGTM

PR #125, opened 2024-11-06
feat: add n_validation_batches option to trainer

JSabadin:
## Efficient Model Training Without Full Test/Validation and Test/Validation Split

If you want to just train the model and export it without extensive validation or testing, you can now achieve this more efficiently using the following configuration:

- **Skip Full Validation**: By setting `trainer.validation_interval` to be the same as `trainer.epochs` and using the same data for training, validation, and testing, you can minimize the validation overhead.
- **Fast Validation Setup**: Specify `n_validation_batches` with a small number to limit the number of validation batches, speeding up the process while ensuring checkpoints are saved correctly for export.
- **Deterministic Loaders**: Both validation and test DataLoaders are now fully deterministic, ensuring reproducibility.

### Example Configuration
```yaml

loader:
  train_view: train
  val_view: train
  test_view: train
  ...
 
trainer:
  validation_interval: *epochs
  n_validation_batches: 2  # Use a small number for faster validation

### DIFF
--- a/luxonis_train/config/config.py
+++ b/luxonis_train/config/config.py
@@ -357,7 +357,7 @@ class TrainerConfig(BaseModelExtraForbid):
     verbose: bool = True
 
     seed: int | None = None
-    n_validation_batches: int | None = None
+    n_validation_batches: PositiveInt | None = None
     deterministic: bool | Literal["warn"] | None = None
     batch_size: PositiveInt = 32
     accumulate_grad_batches: PositiveInt = 1
--- a/luxonis_train/config/config.py
+++ b/luxonis_train/config/config.py
@@ -357,6 +357,7 @@ class TrainerConfig(BaseModelExtraForbid):
     verbose: bool = True
 
     seed: int | None = None
+    n_validation_batches: int | None = None
     deterministic: bool | Literal["warn"] | None = None
     batch_size: PositiveInt = 32
     accumulate_grad_batches: PositiveInt = 1
--- a/luxonis_train/core/core.py
+++ b/luxonis_train/core/core.py
@@ -179,9 +179,9 @@ class LuxonisModel:
                     indices = list(
                         range(min(n_samples, len(self.loaders[view])))
                     )
-                    self.loaders[view] = torch_data.Subset(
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: luxonis_train/config/config.py [hunk 1]
```diff
@@ -357,6 +357,7 @@ class TrainerConfig(BaseModelExtraForbid):
     verbose: bool = True
 
     seed: int | None = None
+    n_validation_batches: int | None = None
```
klemen1999, line 360: Let's make this `PositiveInt | None = None`

luxonis_train/core/core.py
```diff
@@ -167,6 +167,22 @@ def __init__(
                 "Weighted sampler is not implemented yet."
             )
 
+        generator = torch.Generator()
```
klemen1999, line 170: Can put these two lines inside the `if`

luxonis_train/core/core.py
```diff
@@ -167,6 +167,22 @@ def __init__(
...
+        generator.manual_seed(self.cfg.trainer.seed or 41)
```
klemen1999, line 171: Let's respect the legacy and put this to 42 if not set :smile:

luxonis_train/core/core.py
```diff
@@ -181,6 +197,7 @@ def __init__(
                 ),
                 pin_memory=self.cfg.trainer.pin_memory,
                 sampler=sampler if view == "train" else None,
+                generator=generator if view in ["val", "test"] else None,
```
klemen1999, line 200: We want to be very careful here. Generator should only be not None if `self.cfg.trainer.n_validation_batches` is not None and view is `val` or `test`. Current code I believe sets the generator always which would mean that val and test set would be always deterministic (since we manually set the random seed).

klemen1999 (commented):
Generally LGTM, left some comments. Also let's add this to parameter to the README. And let's think about testing it: We don't need to really train, just init the LuxonisModel and check that the length of the loader is correct when we don't and when we do set this parameter.

kozlov721 (approved):
LGTM

PR #128, opened 2024-11-07
Pretrained weights for detection-light and segmentation-light

klemen1999:
## Pretrained weights for detection-light and segmentation-light

This PR adds option to download pretrained weights for detection and segmentation predefined models (light versions). Checkpoints are split per node and saved in the release in luxonis-train ([v0.1.0](https://github.com/luxonis/luxonis-train/releases/tag/v0.1.0-beta)). If they are not present at the start of the training then we automatically download them otherwise we used the cached ones. 
By default we set `download_weights` to `True` for all backbones (where applicable) and if using predefined models we also download them for neck and head nodes.

### DIFF
--- a/configs/detection_light_model.yaml
+++ b/configs/detection_light_model.yaml
@@ -1,4 +1,5 @@
 # Example configuration for training a predefined light detection model
+# NOTE: This example downloads pretrained COCO weights and trainig parameters are already prepared for fine tuning
 
 model:
   name: detection_light
--- a/configs/detection_light_model.yaml
+++ b/configs/detection_light_model.yaml
@@ -1,5 +1,5 @@
 # Example configuration for training a predefined light detection model
-# NOTE: This example downloads pretrained COCO weights and trainig parameters are already prepared for fine tuning
+# NOTE: This example downloads pretrained COCO weights and training parameters are already prepared for fine tuning
 
 model:
   name: detection_light
--- a/configs/detection_light_model.yaml
+++ b/configs/detection_light_model.yaml
@@ -31,15 +31,21 @@ trainer:
   optimizer:
     name: SGD
     params:
-      lr: 0.02
-      momentum: 0.937
-      weight_decay: 0.0005
-      dampening: 0.0
-      nesterov: true
+      lr: 0.0032           
+      momentum: 0.843      
+      weight_decay: 0.00036 
+      dampening: 0.0        
+      nesterov: true    
 
   scheduler:
     name: CosineAnnealingLR
     params:
-      T_max: *epochs
-      eta_min: 0.0002
-      last_epoch: -1
+      T_max: *epochs         
+      eta_min: 0.000384
+
+  callbacks:
+    - name: EMACallback
+      params:
+        decay: 0.9999 
+        use_dynamic_decay: True 
+        decay_tau: 2000
--- a/luxonis_train/config/predefined_models/detection_model.py
+++ b/luxonis_train/config/predefined_models/detection_model.py
@@ -18,6 +18,8 @@ VariantLiteral: TypeAlias = Literal["light", "heavy"]
 class DetectionVariant(BaseModel):
     backbone: str
     backbone_params: Params
+    neck_params: Params
+    head_params: Params
 
 
 def get_variant(variant: VariantLiteral) -> DetectionVariant:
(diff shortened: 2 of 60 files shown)


luxonis_train/config/predefined_models/detection_model.py
```diff
@@ -27,10 +29,14 @@ def get_variant(variant: VariantLiteral) -> DetectionVariant:
...
             backbone="EfficientRep",
             backbone_params={"variant": "n"},
+            neck_params={"variant": "n", "download_weights": True},
+            head_params={"download_weights": True},
         ),
         "heavy": DetectionVariant(
             backbone="EfficientRep",
             backbone_params={"variant": "l"},
+            neck_params={"variant": "l", "download_weights": True},
+            head_params={"download_weights": False},
```
sokovninn, line 39: We have a pretrained backbone for the heavy detection model, but not a pretrained head, right?

configs/detection_light_model.yaml
```diff
@@ -31,15 +32,21 @@ trainer:
   optimizer:
     name: SGD
     params:
-      lr: 0.02
-      momentum: 0.937
-      weight_decay: 0.0005
-      dampening: 0.0
-      nesterov: true
+      lr: 0.0032
```
kozlov721, line 35: Just out of curiosity where do these values come from?

sokovninn (approved):
LGTM, left a few comments.

luxonis_train/config/predefined_models/detection_model.py
klemen1999: Not yet. We've only trained light detection model on COCO so we only have "nano" variant for the backbone, neck and head. We'll be adding more gradually, as we go.

luxonis_train/nodes/README.md
```diff
@@ -53,7 +53,7 @@ Adapted from [here](https://pytorch.org/vision/main/models/resnet.html).
 | Key                | Type                                      | Default value | Description                            |
 | ------------------ | ----------------------------------------- | ------------- | -------------------------------------- |
 | `variant`          | `Literal["18", "34", "50", "101", "152"]` | `"18"`        | Variant of the network                 |
-| `download_weights` | `bool`                                    | `False`       | If True download weights from ImageNet |
+| `download_weights` | `bool`                                    | `True`        | If True download weights from ImageNet |
```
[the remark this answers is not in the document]
klemen1999: Yes eventually, specially for the "whole models" (e.g. predefined detection or predefined segmentation model"

configs/detection_light_model.yaml
```diff
@@ -1,4 +1,5 @@
 # Example configuration for training a predefined light detection model
+# NOTE: This example downloads pretrained COCO weights and training parameters are already prepared for fine tuning
```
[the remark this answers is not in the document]
klemen1999: Currently we allow loading predefined weights for specific module (e.g. our COCO weights for EfficientRep) and we also allow loading custom weights for the whole architecture through `model.weights` in YAML. 
IMO it's too early to allow loading specific checkpoints for each Node - by too early I mean that I don't see a use case for it. But if there is we can add this as well quite easily.

luxonis_train/utils/general.py
```diff
@@ -139,3 +143,63 @@ def get_with_default(
...
+                if i == retry:
+                    logger.warning("Download failed, retry limit reached.")
+                    return None
+                logger.warning(f"Download failed, retrying {i+1}/{retry} ...")
+
+
+def clean_url(url: str) -> str:
+    """Strip auth from URL, i.e. https://url.com/file.txt?auth -> https://url.com/file.txt."""
+    url = (
+        Path(url).as_posix().replace(":/", "://")
```
kozlov721, line 196: You can use `PurePosixPath` instead of `Path` here so you don't have to deal with Windows

kozlov721 (approved):
Left a small suggestion, otherwise LGTM

configs/detection_light_model.yaml
klemen1999: Official YoloV6 finetune config

luxonis_train/config/predefined_models/detection_model.py
sokovninn: Ok, but why do we have `"download_weights": True` here in the heavy variant?

klemen1999: Oh yeah nice catch, will fix.

luxonis_train/config/predefined_models/detection_model.py
```diff
@@ -27,10 +29,14 @@ def get_variant(variant: VariantLiteral) -> DetectionVariant:
         "light": DetectionVariant(
             backbone="EfficientRep",
             backbone_params={"variant": "n"},
+            neck_params={"variant": "n", "download_weights": True},
+            head_params={"download_weights": True},
         ),
         "heavy": DetectionVariant(
             backbone="EfficientRep",
             backbone_params={"variant": "l"},
+            neck_params={"variant": "l", "download_weights": True},
```
klemen1999, line 38:
```suggestion
            neck_params={"variant": "l", "download_weights": False},
```

PR #129, opened 2024-11-08
add test for cli commands

JSabadin:
# Add CLI Command Tests

### Summary
This PR introduces tests for CLI commands. The tests cover both successful and failure scenarios.

### DIFF
--- a/tests/integration/test_cli_commands.py
+++ b/tests/integration/test_cli_commands.py
@@ -3,6 +3,7 @@ import os
 import subprocess
 import sys
 from pathlib import Path
+from typing import Tuple
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


tests/integration/test_cli_commands.py
```diff
@@ -0,0 +1,68 @@
...
+
+import pytest
+
+os.environ["PYTHONIOENCODING"] = "utf-8"
+
+sys.stdout.reconfigure(encoding="utf-8")
+sys.stderr.reconfigure(encoding="utf-8")
+
+ONNX_PATH = Path("tests/integration/model.onnx")
+ARCHIVE_PATH = Path("tests/integration/model.nn")
```
klemen1999, line 16: This doesn't look like it is used anywhere

klemen1999 (approved):
LGTM

tests/integration/test_cli_commands.py
```diff
@@ -0,0 +1,66 @@
...
+os.environ["PYTHONIOENCODING"] = "utf-8"
+
+sys.stdout.reconfigure(encoding="utf-8")  # type: ignore
+sys.stderr.reconfigure(encoding="utf-8")  # type: ignore
+
+ONNX_PATH = Path("tests/integration/model.onnx")
+
+
+@pytest.fixture(scope="session", autouse=True)
+def clear_files() -> Generator[None, None, None]:
```
kozlov721: Just a note; it's fine not to add return types to the fixtures, they work by magic and adding the hint doesn't really help ,uch (editors/type checkers have no use for it and the generator type is also not very readable for humans)

PR #132, opened 2024-11-14
Custom optimizer

JSabadin:
# Custom SGD Optimizer Proposal

This PR introduces a custom implementation of the SGD optimizer with a unique step mechanism that is crucial for optimal performance. 

Here is the YAML configuration file that yields the best results using this custom training strategy:
```yaml
model:
  nodes:
    - name: EfficientRep
      params:
        variant: "n"
        download_weights: False

    - name: RepPANNeck
      params:
        download_weights: False
      inputs:
        - EfficientRep

    - name: EfficientBBoxHead
      params:
        download_weights: False
      inputs:
        - RepPANNeck

  losses:
    - name: AdaptiveDetectionLoss
      attached_to: EfficientBBoxHead
      params:
        iou_type: "siou"
        n_warmup_epochs: 0
        iou_loss_weight: 20 # Should be 2.5 * accumulate_grad_batches
        class_loss_weight: 8 # Should be 1 * accumulate_grad_batches
        

  metrics:
    - name: MeanAveragePrecision
      is_main_metric: true
      attached_to: EfficientBBoxHead

  visualizers:
    - name: BBoxVisualizer
      attached_to: EfficientBBoxHead


loader:
  name: LuxonisLoaderTorch

  train_view: train
  val_view: train
  test_view: train

  params:
    dataset_name: cow_finetune
    bucket_storage: "local"

trainer:
  seed: 42
  precision: "16-mixed"
  preprocessing:
    train_image_size: [640, 640]
    keep_aspect_ratio: true
    normalize:
      active: true
      params:
        mean: [0., 0., 0.]
        std: [1, 1, 1]

  skip_last_batch: False
  batch_size: 8
  accumulate_grad_batches: 8
  epochs: &epochs 80
  num_workers: 4
  validation_interval: 10
  num_log_images: 8
  n_sanity_val_steps: 0

  training_strategy:
    name: "TripleLRSGDStrategy"
    params: 
      warmup_epochs: 3
      warmup_bias_lr: 0.1
      warmup_momentum: 0.8
      lr: 0.02
      lre: 0.0002
      momentum: 0.937     
      weight_decay: 0.0005     
      nesterov: True


  callbacks:
    - name: EMACallback
      params:
        decay: 0.9999 
        use_dynamic_decay: True 
        decay_tau: 2000

### CODE REVIEW

**Comment by JSabadin** [comment]
I can create CustomScheduler class:
`class CustomScheduler:
    def __init__(self, optimizer, cfg_trainer, max_stepnum):
`
This class will also have a method `update_learning_rate(self, current_epoch) `to update the learning rate each time the `on_after_backward `hook is called in PyTorch Lightning.

Similarly, a custom optimizer can be created using the CustomOptimizer class:
`class CustomOptimizer:
    def __init__(self, model, cfg_optimizer):
`
This approach will make the optimizer configuration in configure_optimizers cleaner and more modular.

File: configs/README.md [hunk 1]
```diff
@@ -376,6 +376,37 @@ trainer:
       eta_min: 0
 ```
 
+### Training Strategy
```
klemen1999, line 379: Also add this section to top "Table of contents"

luxonis_train/models/luxonis_lightning.py
```diff
@@ -268,6 +273,16 @@ def __init__(
 
         self.load_checkpoint(self.cfg.model.weights)
 
+        if self.cfg.trainer.training_strategy.params:
```
klemen1999, line 276: Check if params are specified to determine if training strategy is specified feels a bit weird. Should the whole training strategy section in config rather be None by default and then if used we make `name` a required parameter like we have for `CallbackConfig`? CC: @kozlov721 also for thoughts

luxonis_train/strategies/base_strategy.py
```diff
@@ -0,0 +1,28 @@
...
+
+import pytorch_lightning as pl
+from luxonis_ml.utils.registry import AutoRegisterMeta
+from torch.optim import Optimizer
+from torch.optim.lr_scheduler import LRScheduler
+
+from luxonis_train.utils.registry import STRATEGIES
+
+
+class BaseTrainingStrategy(
```
klemen1999, line 11: Let's add this to "Customizations" docs as we now allow registering custom strategies

luxonis_train/models/luxonis_lightning.py
JSabadin: Yeah, I completely agree—that approach would make much more sense. Initially, I followed the logic of using predefined Adam and constantLR for the optimizer and scheduler, applying the same idea to the strategy. However, that feels unintuitive and should definitely be changed. Also, @kozlov721, what’s your take on this?

JSabadin: Now I see what I was doing wrong (when getting that error). I'll fix it...

PR #133, opened 2024-11-18
FOMO detection head

JSabadin:
# FOMO Detection Head

Introducing a new FOMO detection head designed for high-speed object detection, focusing specifically on detecting the centers of objects. The model's heatmap output is directly converted into object detections, streamlining the detection process.

The training process involves extracting the center of each bounding box and using this information to generate the target heatmap. The loss function used is binary cross entropy with logits.

Here is an example of overfitting:  
fomo


tersekmatija:
Thanks, can we try:
- Training where backbone is frozen and only head is trained? What speed-ups do we achieve?
- In post processing, can we look at the heatmaps to extra full object size?

klemen1999 (commented):
Generally LGTM, things to also add:
- predefined FOMO model with e.g. 2 variants
- Docs
- Add to predefined model tests

luxonis_train/nodes/heads/fomo_head.py
```diff
@@ -0,0 +1,94 @@
...
+        heatmap = self.conv3(x)
+        return heatmap
+
+    def wrap(self, heatmap: Tensor) -> Packet[Tensor]:
+        if self.training or self.export:
+            return {
+                "features": [heatmap],
+            }
+        else:
+            boxes = self._heatmap_to_xyxy(heatmap)
```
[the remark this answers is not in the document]
JSabadin: No, it’s not compatible with mAP metric. Its purpose is only to detect and classify the object, not to get the entire bounding box. So for each object, only its approximate center will be obtained.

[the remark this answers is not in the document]
JSabadin: In that case, a method to cluster the data is needed. Additionally, the labels would not be the centroids but rather Gaussian-smoothed centers. The challenge is that overlapping objects will complicate things significantly, as the method cannot detect overlapping objects. If objects overlap, only one will be detected, the information about the locations of the other objects will be lost quicker, compared to only detecting the points.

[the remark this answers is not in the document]
JSabadin: > In post processing, can we look at the heatmaps to extra full object size?

CC: @tersekmatija also here.

configs/detection_fomo_heavy_model.yaml
```diff
@@ -0,0 +1,42 @@
+# Example configuration for training a predefined fomo heavy detection model
+# NOTE: This example downloads pretrained COCO weights and training parameters are already prepared for fine tuning
+
+model:
+  name: detection_fomo
+  predefined_model:
+    name: DetectionModel
```
klemen1999, line 7: Let's call it FOMOModel (same for the other config)

configs/detection_fomo_heavy_model.yaml
```diff
@@ -0,0 +1,42 @@
+# Example configuration for training a predefined fomo heavy detection model
+# NOTE: This example downloads pretrained COCO weights and training parameters are already prepared for fine tuning
```
klemen1999, line 2: MobileNetv2 weights are from ImageNet ([source](https://pytorch.org/vision/main/models/generated/torchvision.models.mobilenet_v2.html#torchvision.models.MobileNet_V2_Weights)) - same for other config

luxonis_train/attached_modules/losses/fomo_localization_loss.py
```diff
@@ -0,0 +1,71 @@
...
+logger = logging.getLogger(__name__)
+
+
+class FOMOLocalizationLoss(BaseLoss[Tensor, Tensor]):
+    node: FOMOHead
+    supported_tasks: list[tuple[TaskType, ...]] = [
+        (TaskType.BOUNDINGBOX, TaskType.KEYPOINTS)
+    ]
+
+    def __init__(self, object_weight: float = 1000, **kwargs: Any):
```
klemen1999, line 23: Original implementation has object_weight=100 as default. Does that not work as well?

tests/integration/test_simple.py
```diff
@@ -73,10 +75,11 @@ def test_predefined_models(
             if "classification" in config_file
             else coco_dataset.dataset_name
         ),
+        "trainer.epochs": 1,
     }
     model = LuxonisModel(config_file, opts)
     model.train()
-    model.test()
+    model.test(view="train")
```
klemen1999, line 82: Is this needed?

configs/detection_fomo_heavy_model.yaml
```diff
@@ -0,0 +1,42 @@
+# Example configuration for training a predefined fomo heavy detection model
```
klemen1999, line 1: Add this model into README [here](https://github.com/luxonis/luxonis-train/blob/main/luxonis_train/config/predefined_models/README.md)

klemen1999 (commented):
Generally LGTM, just left couple of comments

tests/integration/test_simple.py
JSabadin: No, it is not needed

luxonis_train/attached_modules/losses/fomo_localization_loss.py
klemen1999: Feel free to set to the value that works best

klemen1999 (approved):

luxonis_train/nodes/heads/fomo_head.py
```diff
@@ -0,0 +1,106 @@
...
+
+from luxonis_train.enums import TaskType
+from luxonis_train.nodes.base_node import BaseNode
+from luxonis_train.utils import Packet
+
+logger = logging.getLogger(__name__)
+
+
+class FOMOHead(BaseNode[list[Tensor], list[Tensor]]):
+    tasks: list[TaskType] = [TaskType.KEYPOINTS, TaskType.BOUNDINGBOX]
```
kozlov721, line 15: Similar to the comment above; I'm a bit confused about the output of the model. It predicts a heatmap that's converted to keypoints that somehow define a bounding box of the center of the object? That's how it looks to me in the description image.
If that's the case, we should stick to only one of those tasks (keypoints would probably make the most sense).

kozlov721 (commented):
LGTM in general, just left some comments about the tasks.
 
Also, would it make sense to add a visualizer for the heatmaps?

luxonis_train/attached_modules/losses/fomo_localization_loss.py
```diff
@@ -0,0 +1,71 @@
...
+
+from .base_loss import BaseLoss
+
+logger = logging.getLogger(__name__)
+
+
+class FOMOLocalizationLoss(BaseLoss[Tensor, Tensor]):
+    node: FOMOHead
+    supported_tasks: list[tuple[TaskType, ...]] = [
+        (TaskType.BOUNDINGBOX, TaskType.KEYPOINTS)
```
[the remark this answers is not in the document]
klemen1999: Because of easier integration with existing code (mainly metrics e.g. OKS) this model requires both bboxes + keypoint at the center of the box. This is because GT labels in the metrics are automatically pulled from label dict and we then need those to be present.

luxonis_train/nodes/heads/fomo_head.py
JSabadin: The issue is that for the metric to function correctly, we need the ground truth bounding boxes to calculate the OKS metric. That is why the dataset must include bounding boxes with a center keypoint for each bounding box.

luxonis_train/attached_modules/losses/fomo_localization_loss.py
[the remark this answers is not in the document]
klemen1999: Yeah, certanly a point where we can thing about improving the internal design because this is indeed kind of a hack. You can prepare the predictions for attached modules using wrap() but perhaps you should be also able to access the GT that get's fed into the attached modules. In most cases you would leave as is but for cases like this you could add the "bounding box" GTs at runtime so it is compatible with metrics. But there also might be some problems with defining required labels etc. this way.

klemen1999 (approved):

PR #139, opened 2024-11-26
Refactor Archiver and add BaseHead

sokovninn:
This PR introduces the `BaseHead` class to standardize head definitions in the NN architecture. The new class enforces required fields like `parser` and provides a structured way to define head-specific metadata via `get_custom_head_config`.  

Additionally, the archiver has been refactored to utilize the `BaseHead` class for generating configurations, reducing redundancy and improving scalability when adding new architectures. This ensures auxiliary heads and other metadata are handled consistently.  

**Key Changes:**  
- Added `BaseHead` class with a `parser` field.  
- Introduced `_get_base_head_config` and `get_custom_head_config` for better separation of base and custom logic.  
- Updated the archiver to use `BaseHead` instances instead of relying solely on the config file.  
- Improved error handling for missing or invalid head configurations.


luxonis_train/nodes/heads/base_head.py
```diff
@@ -0,0 +1,77 @@
...
+    is_softmaxed: bool | None = None
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+    def __init_subclass__(cls, **kwargs):
+        """Check if head has all required attributes."""
+        super().__init_subclass__()
+        if cls.parser is None:
+            raise ValueError(f"Head `{cls.__name__}` must define a parser.")
```
sokovninn, line 42: Should we enforce specifying the parser here or it could be `None`? For example, we don't have a defined parser for `DiscSubNetHead`.

kozlov721: I don't think we should enforce it. The NN archive support imho should be opt-in for new heads because it's only relevant if the user wants to use the model with dai. 

I think we should leave it `None` as default and also remove the `@abstractmethod` from `_get_custom_head_config`. Instead, it would raise `NotImplementedError` by default. In the `Core.archive` you'd check whether this method raises or not.
If it does, the archiving fails with a message like "Archiving failed, node {node_name} does not implement ...".
If it doesn't raise, only then we'd check that `parser` and `is_softmaxed` are not `None`.

luxonis_train/nodes/heads/base_head.py
```diff
@@ -0,0 +1,77 @@
...
+            "parser": self.parser,
+            "metadata": {
+                "is_softmax": self.is_softmaxed,
+                "classes": self.class_names,
+                "n_classes": self.n_classes,
+            },
+        }
+
+    @abstractmethod
+    def _get_custom_head_config(self) -> dict:
```
kozlov721, line 71: Also we should probably make it public if it's intended to be overriden.

luxonis_train/nodes/heads/base_head.py
```diff
@@ -0,0 +1,77 @@
...
+    """Base class for all heads in the model.
+
+    @type parser: str | None
+    @ivar parser: Parser to use for the head.
+    @type is_softmaxed: bool | None
+    @ivar is_softmaxed: Whether the head uses softmax or not.
+    """
+
+    parser: (
+        Literal[
```
klemen1999, line 25: Let's make this of type `str | None`  to allow someone to register a custom head with a  custom named parser which they would implement themselves.

luxonis_train/nodes/heads/base_head.py
```diff
@@ -0,0 +1,77 @@
...
+            "ClassificationParser",
+            "YOLO",
+            "YoloDetectionNetwork",
+            "SegmentationParser",
+        ]
+        | None
+    ) = None
+    is_softmaxed: bool | None = None
```
klemen1999, line 33: IMO this should this be specifically added only to the heads where relevat (as we do for example with `iou_threshold` on detection head)

luxonis_train/nodes/heads/efficient_keypoint_bbox_head.py
```diff
@@ -214,3 +215,16 @@ def _process_to_bbox_and_kps(
...
+
+        @rtype: dict
+        @return: Custom head configuration.
+        """
+        return {
+            "iou_threshold": self.iou_thres,
+            "conf_threshold": self.conf_thres,
+            "max_det": self.max_det,
+            "n_keypoints": self.n_keypoints,
+        }
```
klemen1999, line 230: Let's also add:
`"subtype": "yolov6"`

Generally these parameters should match the ones specifid in the respective parser in [depthai-nodes](https://github.com/luxonis/depthai-nodes/tree/main/depthai_nodes/ml/parsers)

luxonis_train/nodes/heads/bisenet_head.py
```diff
@@ -56,3 +57,13 @@ def forward(self, inputs: Tensor) -> Tensor:
...
+
+    def get_custom_head_config(self) -> dict:
+        """Returns custom head configuration.
+
+        @rtype: dict
+        @return: Custom head configuration.
+        """
+        return {
+            "is_softmax": False,
+        }
```
klemen1999, line 69: Segmentation parser in DAI-nodes needs "output_layer_name" and "classes_in_one_layer" - [reference](https://github.com/luxonis/depthai-nodes/blob/main/depthai_nodes/ml/parsers/segmentation.py)
(same for ddnet and segmentation head)

klemen1999, line 69: Segmentation parser in DAI-nodes needs "output_layer_name" and "classes_in_one_layer" - [reference](https://github.com/luxonis/depthai-nodes/blob/main/depthai_nodes/ml/parsers/segmentation.py)
(same for ddnet and segmentation head)

luxonis_train/nodes/heads/classification_head.py
```diff
@@ -31,3 +32,13 @@ def __init__(self, dropout_rate: float = 0.2, **kwargs: Any):
...
+
+    def get_custom_head_config(self) -> dict:
+        """Returns custom head configuration.
+
+        @rtype: dict
+        @return: Custom head configuration.
+        """
+        return {
+            "is_softmax": False,
+        }
```
klemen1999, line 44: Also needs "output_layer_name" - [reference](https://github.com/luxonis/depthai-nodes/blob/main/depthai_nodes/ml/parsers/classification.py)

klemen1999, line 44: Also needs "output_layer_name" - [reference](https://github.com/luxonis/depthai-nodes/blob/main/depthai_nodes/ml/parsers/classification.py)

luxonis_train/nodes/heads/discsubnet_head/discsubnet_head.py
```diff
@@ -109,3 +109,11 @@ def wrap(
...
                 "segmentation": [seg_out],
             }
+
+    def get_custom_head_config(self) -> dict:
+        """Returns custom head configuration.
+
+        @rtype: dict
+        @return: Custom head configuration.
+        """
+        return {}
```
klemen1999, line 119: AFAIK we should be able to use `SegmentationParser` here as well. CC: @JSabadin  on the correct parameters so this model can work with the dai-nodes parser

klemen1999 (commented):
Generally LGTM, we just need to ensure that the end archives can be seamlessly then used with parsers in dai-nodes

klemen1999 (commented):
Generally LGTM, we just need to ensure that the end archives can be seamlessly then used with parsers in dai-nodes

luxonis_train/nodes/heads/efficient_keypoint_bbox_head.py
```diff
@@ -17,6 +17,7 @@
 
 class EfficientKeypointBBoxHead(EfficientBBoxHead):
     tasks: list[TaskType] = [TaskType.KEYPOINTS, TaskType.BOUNDINGBOX]
+    parser: str = "YoloDetectionNetwork"
```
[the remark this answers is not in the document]
sokovninn: I took this value from the `ImplementedHeads` enum. Do we have a new parser for this head?

luxonis_train/nodes/heads/bisenet_head.py
sokovninn: Isn't this info already contained in the `outputs` filed in the generated config?

klemen1999: Yeah correct. Disregard my comment about "output_layer_name", it gets pulled from "outputs"

luxonis_train/nodes/heads/efficient_keypoint_bbox_head.py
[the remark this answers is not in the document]
klemen1999: Yeah those were not updated with the changes in dai-nodes but they should be now as we nailed down the parser names.

luxonis_train/nodes/heads/discsubnet_head/discsubnet_head.py
JSabadin: The default parameters are the correct ones.

luxonis_train/nodes/heads/base_head.py
```diff
@@ -0,0 +1,58 @@
...
+):
+    """Base class for all heads in the model.
+
+    @type parser: str | None
+    @ivar parser: Parser to use for the head.
+    """
+
+    parser: str | None = None
+
+    def __init__(self, **kwargs):
```
kozlov721, line 22: The `__init__` isn't necessary

luxonis_train/nodes/heads/base_head.py
```diff
@@ -0,0 +1,58 @@
...
+from luxonis_train.nodes.base_node import (
+    BaseNode,
+    ForwardInputT,
+    ForwardOutputT,
+)
+
+
+class BaseHead(
+    BaseNode[ForwardInputT, ForwardOutputT],
+    Generic[ForwardInputT, ForwardOutputT],
```
kozlov721, line 12: I think the `Generic` shouldn't be necessary

luxonis_train/utils/general.py
```diff
@@ -203,3 +203,15 @@ def clean_url(url: str) -> str:
 def url2file(url: str) -> str:
     """Convert URL to filename, i.e. https://url.com/file.txt?auth -> file.txt."""
     return Path(clean_url(url)).name
+
+
+def deep_merge_dicts(dict1: dict, dict2: dict) -> None:
```
kozlov721, line 208: We should add unittests for the deep merge so it has 100 % coverage

kozlov721 (approved):
LGTM, left some small comments

PR #141, opened 2024-12-06
Re-Identification Model

CaptainTrojan:
Original [CU Task](https://app.clickup.com/t/86c16994c)

Added the GFNv2 backbone for use in re-ID tasks. Added integration with pytorch-metric-learning for ~30 loss functions (customizable) and two custom reID/embedding metrics. Implemented a test to verify trainability, exportability and inference.


luxonis_train/attached_modules/losses/__init__.py
```diff
@@ -7,6 +7,7 @@
 from .ohem_bce_with_logits import OHEMBCEWithLogitsLoss
 from .ohem_cross_entropy import OHEMCrossEntropyLoss
 from .ohem_loss import OHEMLoss
+from .pml_loss import MetricLearningLoss
```
klemen1999, line 10: What does"pml" stands for?

luxonis_train/attached_modules/losses/pml_loss.py
```diff
@@ -0,0 +1,122 @@
...
+        embedding_size: int = 512,
+        cross_batch_memory_size=0,
+        loss_kwargs: dict | None = None,
+        *args,
+        **kwargs,
+    ):
+        super().__init__(*args, **kwargs)
+        if loss_kwargs is None:
+            loss_kwargs = {}
+        self.loss_func = loss_dict[loss_name](
```
klemen1999, line 92: if loss name is 1:1 mapping to actual class name then you don't need `loss_dict`, you can do e.g.
```
try:
    eval(loss_name)
except:
    # loss name not correct
```

luxonis_train/attached_modules/losses/pml_loss.py
```diff
@@ -0,0 +1,122 @@
...
+    "SoftTripleLoss": SoftTripleLoss,
+    "SphereFaceLoss": SphereFaceLoss,
+    "SubCenterArcFaceLoss": SubCenterArcFaceLoss,
+    "SupConLoss": SupConLoss,
+    "TripletMarginLoss": TripletMarginLoss,
+    "TupletMarginLoss": TupletMarginLoss,
+}
+
+
+class MetricLearningLoss(BaseLoss):
```
klemen1999, line 79: All the losses here are for embedding task right? And this class is a wrapper for all of them? Maybe we change the name so it reflects this more e.g. `EmbeddingLossWrapper` (probably there is a better name)

luxonis_train/attached_modules/losses/pml_loss.py
```diff
@@ -0,0 +1,122 @@
...
+
+        # self.miner_func = miner_func
+
+    def prepare(self, inputs, labels):
+        embeddings = inputs["features"][0]
+
+        assert (
+            labels is not None and "id" in labels
+        ), "ID labels are required for metric learning losses"
+        IDs = labels["id"][0][:, 0]
```
klemen1999, line 114: Nitpick: I'd put the variable name lowercase (e.g. `ids`). If you go with this then also chage other occurances of `IDs`

luxonis_train/attached_modules/losses/pml_loss.py
```diff
@@ -0,0 +1,122 @@
...
+            **loss_kwargs
+        )  # Instantiate the loss object
+        if cross_batch_memory_size > 0:
+            if loss_name in CrossBatchMemory.supported_losses():
+                self.loss_func = CrossBatchMemory(
+                    self.loss_func, embedding_size=embedding_size
+                )
+            else:
+                # Warn that cross_batch_memory_size is ignored
+                warnings.warn(
```
klemen1999, line 102: Let's use logger for any kind of printing or warning

luxonis_train/attached_modules/losses/pml_loss.py
```diff
@@ -0,0 +1,122 @@
...
+                    f"Cross batch memory is not supported for {loss_name}. Ignoring cross_batch_memory_size"
+                )
+
+        # self.miner_func = miner_func
+
+    def prepare(self, inputs, labels):
+        embeddings = inputs["features"][0]
+
+        assert (
```
klemen1999, line 111: Let's do an if check and raise error since asserts can be bypassed if program exectured wth -O flag

luxonis_train/attached_modules/metrics/pml_metrics.py
```diff
@@ -0,0 +1,258 @@
...
+    indices_equal = torch.eye(
+        labels.shape[0], dtype=torch.uint8, device=labels.device
+    )
+    indices_not_equal = ~indices_equal
+    labels_equal = labels.unsqueeze(0) == labels.unsqueeze(1)
+    mask = indices_not_equal & labels_equal
+    return mask
+
+
+class ClosestIsPositiveAccuracy(BaseMetric):
```
klemen1999, line 66: Nitpick: Let's have classes on top and helper functions below. Or if they are used only inside specific class then make them class private functions?

luxonis_train/loaders/utils.py
```diff
@@ -38,6 +38,7 @@ def collate_fn(
             TaskType.CLASSIFICATION,
             TaskType.SEGMENTATION,
             TaskType.ARRAY,
+            TaskType.LABEL,
```
klemen1999, line 41: We'll wait for the new luxonis-ml aug refactor to figure what is the best way to handle this. CC: @kozlov721 on this also (can sync offline on the requirements)

luxonis_train/nodes/backbones/ghostfacenet.py
```diff
@@ -0,0 +1,511 @@
...
+import math
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+
+from luxonis_train.nodes.base_node import BaseNode
+
+
+def _make_divisible(v, divisor, min_value=None):
```
klemen1999, line 13: Check if some of these util functions and blocks are already implemented and can be reused (e.g. ConvBNAct can be replaced with [ConvModule](https://github.com/luxonis/luxonis-train/blob/8be019f7c6924dd902532b8156c09c917eb435b0/luxonis_train/nodes/blocks/blocks.py#L84), then [SqueezeExciteBlock](https://github.com/luxonis/luxonis-train/blob/8be019f7c6924dd902532b8156c09c917eb435b0/luxonis_train/nodes/blocks/blocks.py#L199C7-L199C25) could be used, etc.)

luxonis_train/nodes/backbones/ghostfacenet.py
```diff
@@ -0,0 +1,511 @@
...
+            x = self.bn_dw(x)
+        if self.se is not None:
+            x = self.se(x)
+        x = self.ghost2(x)
+        x += self.shortcut(residual)
+        return x
+
+
+# NODES.register_module()
+class GhostFaceNetsV2(BaseNode[torch.Tensor, list[torch.Tensor]]):
```
klemen1999, line 334: Let's put this on the top of this script as it is the main class

klemen1999 (commented):
Thanks, left some comments

luxonis_train/attached_modules/metrics/pml_metrics.py
```diff
@@ -0,0 +1,258 @@
+import torch
```
kozlov721, line 1: Most of the code here is not covered by tests, please add tests for all the metrics. For new architectures we should require 100% coverage before merge.

luxonis_train/nodes/backbones/ghostfacenet.py
```diff
@@ -0,0 +1,511 @@
...
+        super().__init__(*args, **kwargs)
+
+        inp_shape = kwargs["input_shapes"][0]["features"][0]
+        # spots_shape = kwargs['input_shapes'][0]['features'][1]
+
+        image_size = inp_shape[2]
+        channels = inp_shape[1]
+        if cfgs is None:
+            self.cfgs = [
+                # k, t, c, SE, s
```
kozlov721, line 409: The configuration is not very readable. We should add the `variant` logic the same way it's done for other configurable backbones (check `MicroNet`)

luxonis_train/nodes/backbones/ghostfacenet.py
```diff
@@ -0,0 +1,511 @@
...
+        return [inputs[0]["features"][0]]
+
+    def wrap(self, outputs):
+        return {"features": [outputs]}
+
+    def set_export_mode(self, mode: bool = True):
+        self.export_mode = mode
+        self.train(not mode)
+
+    def __init__(
```
kozlov721, line 345: `__init__` should be moved up

luxonis_train/nodes/backbones/ghostfacenet.py
kozlov721, line 334: Let's make it `Tensor` instead of `torch.Tensor` for consistency with the rest of the codebase.

luxonis_train/nodes/backbones/ghostfacenet.py
```diff
@@ -0,0 +1,511 @@
...
+        if self.se is not None:
+            x = self.se(x)
+        x = self.ghost2(x)
+        x += self.shortcut(residual)
+        return x
+
+
+# NODES.register_module()
+class GhostFaceNetsV2(BaseNode[torch.Tensor, list[torch.Tensor]]):
+    def unwrap(self, inputs):
```
kozlov721, line 335: You don't need to override `unwrap` and `wrap` in this case. The default implementation does all what you need.

luxonis_train/nodes/backbones/ghostfacenet.py
```diff
@@ -0,0 +1,511 @@
...
+        @param bn_epsilon: Batch normalization epsilon. Defaults to 1e-5.
+        @type init_kaiming: bool
+        @param init_kaiming: If True, initializes the weights using the Kaiming initialization. Defaults to True.
+        @type block_args: dict
+        @param block_args: Arguments to pass to the block. Defaults to None.
+        """
+        # kwargs['_tasks'] = {TaskType.LABEL: 'features'}
+        super().__init__(*args, **kwargs)
+
+        inp_shape = kwargs["input_shapes"][0]["features"][0]
```
kozlov721, line 402: You should use `BaseNode.in_sizes` property (or `in_channels` / `in_width` / ... depending on what you need). Additionally you should add class-level attribute annotations for those properties for an automatic type check

```python

class GhostFaceNetsV2(...):
    in_channels: int

    ...
```

kozlov721: You should never need to manually work with the insides of `BaseNode`. As a rule of thumb; if you use `["features"]` in the code, then there's a method / property that does it for you.

luxonis_train/nodes/backbones/ghostfacenet.py
```diff
@@ -0,0 +1,511 @@
...
+
+# NODES.register_module()
+class GhostFaceNetsV2(BaseNode[torch.Tensor, list[torch.Tensor]]):
+    def unwrap(self, inputs):
+        return [inputs[0]["features"][0]]
+
+    def wrap(self, outputs):
+        return {"features": [outputs]}
+
+    def set_export_mode(self, mode: bool = True):
```
kozlov721, line 341: You also don't need to define this manually; the default implementation is enough. (also you're setting non-existent attribute `export_mode`)

luxonis_train/nodes/backbones/ghostfacenet.py
```diff
@@ -0,0 +1,511 @@
...
+    if min_value is None:
+        min_value = divisor
+    new_v = max(min_value, int(v + divisor / 2) // divisor * divisor)
+    # Make sure that round down does not go down by more than 10%.
+    if new_v < 0.9 * v:
+        new_v += divisor
+    return new_v
+
+
+def hard_sigmoid(x, inplace: bool = False):
```
tersekmatija, line 29: Could this be part of some utils, I'd assume this is repeating elsewhere?

luxonis_train/attached_modules/losses/__init__.py
CaptainTrojan: Pytorch Metric Learning

CaptainTrojan:
Thanks very much everyone for the review and suggestions. Incorporated everything in, except for the new luxonis-ml nested label changes, as that should probably be done in a different PR. Thoughts?

CaptainTrojan:
Locally, I'm getting

```
(luxonis) captaintrojan@cpt-debian-PC:~/Projects/luxonis-train$ pre-commit run --show-diff-on-failure --color=always --all-files
ruff.....................................................................Passed
ruff-format..............................................................Passed
docformatter.............................................................Passed
mdformat.................................................................Passed
```

Could it be that I have a wrong version of mdformat?

PR #142, opened 2024-12-09
Better pretrained weight loading

klemen1999:
Future proofing for pretrained weights so we can add them later one. Once available they will be automatically downloaded and used.

### DIFF
--- a/luxonis_train/nodes/backbones/ddrnet/ddrnet.py
+++ b/luxonis_train/nodes/backbones/ddrnet/ddrnet.py
@@ -241,8 +241,13 @@ class DDRNet(BaseNode[Tensor, list[Tensor]]):
         self.layer5_bottleneck_expansion = layer5_bottleneck_expansion
         self.init_params()
 
-        if download_weights and var.weights_path:
-            self.load_checkpoint(var.weights_path)
+        if download_weights:
+            if var.weights_path:
+                self.load_checkpoint(var.weights_path)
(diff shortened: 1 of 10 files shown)


luxonis_train/nodes/backbones/ddrnet/ddrnet.py
```diff
@@ -241,8 +241,13 @@ def __init__(
...
         self.init_params()
 
-        if download_weights and var.weights_path:
-            self.load_checkpoint(var.weights_path)
+        if download_weights:
+            if var.weights_path:
+                self.load_checkpoint(var.weights_path)
+            else:
+                logger.warning(
+                    f"No checkpoint available for {self.__class__.__name__}, skipping."
```
kozlov721, line 249: There's the `BaseNode.name` property that can be used instead of `__class__.__name__`.

kozlov721 (approved):
LGTM, just left a small comment. We could also add some tests for the various `get_variant_weights` functions that could just ping the URL to verify it's correct.

klemen1999:
> We could also add some tests for the various get_variant_weights functions that could just ping the URL to verify it's correct.

I agree, the only issue currently is that most of these URLs don't exist yet. They will be there when release is made and the checkpoints are added to the release. So I'll add these tests after release is made.

kozlov721:
> So I'll add these tests after release is made.

Sounds good

PR #145, opened 2024-12-13
fix: ddp bug

JSabadin:
# PR Description

This PR updates the metric computation step to ensure that all returned metrics (`map` and related values) are moved to the GPU device before logging. Without this change, calls to `pl.log` would trigger runtime errors by attempting distributed reductions (`all_reduce`) on CPU tensors under the `nccl` backend in multi-GPU training environments.

### DIFF
--- a/luxonis_train/attached_modules/metrics/mean_average_precision.py
+++ b/luxonis_train/attached_modules/metrics/mean_average_precision.py
@@ -83,5 +83,7 @@ class MeanAveragePrecision(
                     )
 
         map = metric_dict.pop("map")
-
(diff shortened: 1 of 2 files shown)


klemen1999 (commented):
Will we encounter same issue if using `MeanAveragePrecisionKeypoints` or other custom metrics?

JSabadin:
Yes, if the output metrics are stored on the device's CPU

klemen1999:
Let's test this out and include the fixes for others as well under this PR in that case

kozlov721 (approved):
Lgtm

PR #147, opened 2024-12-20
Pinned down mdormat-gfm version

klemen1999:
Newest version doesn't work correctly so we pinned down to the last compatible, working one.


JSabadin (approved):
LGTM

kozlov721 (approved):

PR #148, opened 2025-01-03
Updated weights loading for DDRNet

klemen1999:
- Updated path for light and heavy predefined segmentation model weights
- Improved how we load DDRNet head weights, previous ones weren't actually taken into account because of wrong values in the state dict


kozlov721 (approved):
LGTM

PR #155, opened 2025-01-21
Fix: Ensure Compatibility of Detection Model Input Shape

JSabadin:
### PR Summary

Added validation for `RepPANNeck` input shapes (image dimensions divisible by 32 and GCD check for height and width of feature maps).

### DIFF
--- a/luxonis_train/nodes/necks/reppan_neck/reppan_neck.py
+++ b/luxonis_train/nodes/necks/reppan_neck/reppan_neck.py
@@ -1,4 +1,5 @@
 import logging
+import math
 from typing import Any, Literal
 
 from torch import Tensor, nn
@@ -71,6 +72,25 @@ class RepPANNeck(BaseNode[list[Tensor], list[Tensor]]):
         """
 
         super().__init__(**kwargs)
+
+        if (
+            self.original_in_shape[-1] % 32 != 0
+            or self.original_in_shape[-2] % 32 != 0
+        ):
+            logger.warning(
+                "Image dimensions should be divisible by 32. This may cause 'RepPANNeck' to crash."
+            )
+
+        if (
+            math.gcd(
+                math.gcd(self.in_width[-1], self.in_width[-2]),
+                self.in_width[-3],
+            )
+            == 1
+        ):
+            raise ValueError(
+                "The greatest common divisor of the input widths of 'RepPANNeck' s 1. Image dimensions should be divisible by 32."
+            )
 
         self.n_heads = n_heads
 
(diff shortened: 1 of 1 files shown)


luxonis_train/nodes/necks/reppan_neck/reppan_neck.py
```diff
@@ -72,6 +73,25 @@ def __init__(
...
 
+        if (
+            self.original_in_shape[-1] % 32 != 0
+            or self.original_in_shape[-2] % 32 != 0
+        ):
+            logger.warning(
+                "Image dimensions should be divisible by 32. This may cause 'RepPANNeck' to crash."
+            )
+
+        if (
```
klemen1999, line 84: Wouldn't a more "logical" check be to iterate from spatially smallest input back and check if W1*2=W2 and H1*2=H2 (we need to also check height also)? And this should start where the neck is attached to (taking into account `attach index`) and be done for `n_heads`.

luxonis_train/nodes/necks/reppan_neck/reppan_neck.py
```diff
@@ -72,6 +72,24 @@ def __init__(
...
+            self.original_in_shape[-1] % 32 != 0
+            or self.original_in_shape[-2] % 32 != 0
+        ):
+            logger.warning(
+                "Image dimensions should be divisible by 32. This may cause 'RepPANNeck' to crash."
+            )
+
+        for i in range(1, n_heads):
+            if (
+                self.in_width[-i] * 2 != self.in_width[-i - 1]
```
klemen1999, line 85: Does `in_width` already take into account attach index? So if we had backbone with 4 outputs and attach index would be on the second to last instead of last output does `in_width[-1]` return width of the second to last output or the last one?

kozlov721: Just to clarify; if the `attach_index` is set to `"all"` or a slice, the `in_height` etc. return a list. If it's se to a single index, then the properties return an int.

kozlov721 (approved):

luxonis_train/nodes/necks/reppan_neck/reppan_neck.py
JSabadin: Yeah, but currently, the slice functionality isn't working due to a bug. When a slice is selected, only the first index is being used. Fortunately, none of the nodes explicitly use something like (0, 1, 2), so this doesn't cause any issues at the moment. However, it should still be fixed. In the case of RepaNeck, 'all' is chosen by default, which works correctly and selects all the outputs from the previous node.

PR #156, opened 2025-01-22
fix ema checkpoint loading

JSabadin:
We already support loading hyperparameters such as `lr_schedulers`, `epoch`, `global_step`, `optimizer_steps`, and `state_dict` (including EMA weights if used) via the `--resume` flag in the `train` command. This PR:
1. Fixes a bug in the EMA implementation.
2. Ensures proper handling of the `on_load_checkpoint` hook, which is called before the `on_fit_start` hook.
3. Introduces a new parameter `trainer.resume_training` to resume training from checkpoints specified in the config file (model.weights).

### DIFF
--- a/luxonis_train/__main__.py
+++ b/luxonis_train/__main__.py
@@ -70,7 +70,7 @@ SourcePathType = Annotated[
 @app.command()
 def train(
     config: ConfigType = None,
-    resume_checkpoint: Annotated[
+    resume_weights: Annotated[
         str | None,
         typer.Option(help="Resume training from this checkpoint."),
     ] = None,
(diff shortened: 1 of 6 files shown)


kozlov721 (approved):

luxonis_train/__main__.py
```diff
@@ -70,7 +70,7 @@ class _ViewType(str, Enum):
 @app.command()
 def train(
     config: ConfigType = None,
-    resume: Annotated[
+    resume_checkpoint: Annotated[
```
klemen1999, line 73: I would rather call it `resume_weights` to be compatible with the config parameters where the parameter is called `weight` (same for paramter name in LuxonisModel .train()

klemen1999 (approved):
LGTM

PR #157, opened 2025-01-23
Fixed docs

kozlov721:
Fixed an error in the docs.


klemen1999 (approved):
LGTM
I assume type checks errors are because of luxonis-ml version mismatch right?

PR #158, opened 2025-01-23
OCR recognition model

sokovninn:
# New OCR recognition model, loss, metric and visualizer

The most important changes are summarized below:

### Losses:
* Introduced `CTCLoss` with optional focal loss weighting in `luxonis_train/attached_modules/losses/ctc_loss.py` and updated `__init__.py` to include `CTCLoss`. [[1]](diffhunk://#diff-6ddbd07c00f0a4de562c44a0a17907bcacca9fab6ebaa043ca2136bcb2d43212R1-R82) [[2]](diffhunk://#diff-ce1717d61204e542f4e8c28966a86704ac17d8e4899709d6a301910cf27e3829R5) [[3]](diffhunk://#diff-ce1717d61204e542f4e8c28966a86704ac17d8e4899709d6a301910cf27e3829R30)
* Updated `luxonis_train/attached_modules/losses/README.md` to document `CTCLoss`.

### Metrics:
* Added `OCRAccuracy` metric for OCR tasks in `luxonis_train/attached_modules/metrics/ocr_accuracy.py` and updated `__init__.py` to include `OCRAccuracy`. [[1]](diffhunk://#diff-edec34826f39559b94c473c6bfa98e4f61d469dec317c95734fdd5cbbafd45aaR1-R100) [[2]](diffhunk://#diff-1c652663b03ea7090307387cfc6caec72593d88e8ff6499e1df8cb177680c1a6R6) [[3]](diffhunk://#diff-1c652663b03ea7090307387cfc6caec72593d88e8ff6499e1df8cb177680c1a6R20)
* Updated `luxonis_train/attached_modules/metrics/README.md` to document `OCRAccuracy`.

### Visualizers:
* Introduced `OCRVisualizer` for visualizing OCR tasks in `luxonis_train/attached_modules/visualizers/ocr_visualizer.py` and updated `__init__.py` to include `OCRVisualizer`. [[1]](diffhunk://#diff-ba0ecae8faecf201412f35efae805e152b7aa4de4b9ec2922e4f2abee9ca9dd7R1-R124) [[2]](diffhunk://#diff-6c436e181c9d30f3a8ac0d78bbe7fb1f89db63d554ef459b49e4c36f34de0ad9R6) [[3]](diffhunk://#diff-6c436e181c9d30f3a8ac0d78bbe7fb1f89db63d554ef459b49e4c36f34de0ad9R27)
* Updated `luxonis_train/attached_modules/visualizers/README.md` to document `OCRVisualizer`.

### Predefined Models:
* Added `OCRRecognitionModel` to `luxonis_train/config/predefined_models/__init__.py` and updated `README.md` to document its components and parameters. [[1]](diffhunk://#diff-950088ee932972eee0e9cc9dda31e6a69a1efc44774652f0349d26b4deb29f13R273-R303) [[2]](diffhunk://#diff-cd64af3e62784a25caad6ae58efe322bcf9a2daf3e8940747401317667ef740eR7) [[3]](diffhunk://#diff-cd64af3e62784a25caad6ae58efe322bcf9a2daf3e8940747401317667ef740eR18)

## Examples from the overfitted model

ocr_recognition-OCRCTCHead_OCRVisualizer_3
ocr_recognition-OCRCTCHead_OCRVisualizer_2
ocr_recognition-OCRCTCHead_OCRVisualizer_1
ocr_recognition-OCRCTCHead_OCRVisualizer_0


sokovninn:
Possible improvements include:

- Adding more advanced OCR metrics
- Adding a temporal [NRTR head](https://github.com/PaddlePaddle/PaddleOCR/blob/main/ppocr/modeling/heads/rec_nrtr_head.py) together with [NRTRLoss](https://github.com/PaddlePaddle/PaddleOCR/blob/main/ppocr/losses/rec_nrtr_loss.py)
- Improving visualization
- Adding a large variant
- Adding a beam search decoder
- Adding OCR-specific augmentations

luxonis_train/nodes/heads/ocr_ctc_head.py
```diff
@@ -0,0 +1,119 @@
...
+
+from torch import Tensor, nn
+from torch.nn import functional as F
+
+from luxonis_train.enums import TaskType
+from luxonis_train.nodes.heads import BaseHead
+from luxonis_train.utils import OCRDecoder, OCREncoder
+
+
+def get_para_bias_attr(l2_decay: float, k: int):
```
klemen1999, line 11: Nitpick: Move this to the bottom so class is on top of the file, IMO cleaner

luxonis_train/nodes/heads/ocr_ctc_head.py
```diff
@@ -0,0 +1,119 @@
...
+        nn.init.uniform_(tensor, -stdv, stdv)
+        tensor.regularizer = l2_decay
+
+    return weight_attr, bias_attr
+
+
+class OCRCTCHead(BaseHead[Tensor, Tensor]):
+    in_channels: int
+    tasks: list[TaskType] = [TaskType.CLASSIFICATION]
+
```
klemen1999, line 28: Does the exported model look the same as PaddleOCR? Can we use the same parser for it from depthai-nodes ([this ](https://github.com/luxonis/depthai-nodes/blob/main/depthai_nodes/ml/parsers/classification_sequence.py)one)? Ideally we want to support the full integration with HubAI

luxonis_train/nodes/necks/svtr_neck/blocks.py
```diff
@@ -0,0 +1,327 @@
...
+            (attn.matmul(v))
+            .permute((0, 2, 1, 3))
+            .reshape((batch_size, -1, self.dim))
+        )
+        x = self.proj(x)
+        x = self.proj_drop(x)
+        return x
+
+
+class Block(nn.Module):
```
klemen1999, line 148: Nitpick: Can we name it something more descriptive?

klemen1999 (commented):
Generally LGTM, left some comments. One thing that we want to also make sure is the integration with HubAI and depthai-nodes - the archived model should have correct archive data so that the parser can parse it.

luxonis_train/nodes/backbones/pplcnet_v3/blocks.py
```diff
@@ -0,0 +1,302 @@
...
+            gamma = branch.weight
+            beta = branch.bias
+            eps = branch.eps
+        assert running_var is not None
+        std = (running_var + eps).sqrt()  # type: ignore
+        t = (gamma / std).reshape((-1, 1, 1, 1)).to(kernel.device)  # type: ignore
+        return kernel * t, beta - running_mean * gamma / std  # type: ignore
+
+
+class SELayer(nn.Module):
```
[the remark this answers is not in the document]
sokovninn: Good point! I didn't notice that they were the same.

configs/ocr_recognition_light_model.yaml
```diff
@@ -0,0 +1,43 @@
+# Example configuration for training a predefined light OCR recognition model
+
+model:
+  name: ocr_recognition_light
+  predefined_model:
+    name: OCRRecognitionModel
+    params:
+      variant: light
+      alphabet: ['a', 'b', 'c', 'd', 'e', 'h', 'l', 'm', 'n', 'o', 'p', 'x', 'y', 'z']
```
kozlov721, line 9: We could have a set of predefined alphabets, so users don't have to enumerate all the letters. Something like `"ascii"`, `"numbers"`, `"alphanum"`, or `"english"` (and possibly some other languages)

kozlov721 (approved):
In general LGTM, left some small comments

luxonis_train/attached_modules/metrics/ocr_accuracy.py
```diff
@@ -0,0 +1,109 @@
...
+        @rtype: tuple[Tensor, Tensor]
+        @return: A tuple of predictions and targets.
+        """
+
+        preds = inputs["/classification"][0]
+        targets_batch = labels["/metadata/text"]
+        target_lengths = labels["/metadata/text_length"].int()
+
+        targets = prepare_batch_targets(targets_batch, target_lengths)
+        targets = self.node.encoder(targets).to(preds.device)  # type: ignore
```
kozlov721, line 58: Instead of `# type: ignore`, you can add a class attribute annotation `node: OCRTHead`. This will make the types work and also it will check on the background that the module is connected to a compatible node.

This applies to the visualizer and loss as well.

luxonis_train/loaders/utils.py
```diff
@@ -45,7 +45,7 @@ def collate_fn(
                 label_box.append(new_ann)
             out_labels[task] = torch.cat(label_box, 0)
 
-        elif task_type == "instance_segmentation":
+        elif task_type == "instance_segmentation" or "metadata" in task_type:
```
kozlov721, line 48: We were discussing privately how to best collate the text labels. Since they can be different lengths, we cannot simply stack them, leaving us with 2 options:
1. Concatenating them (as it's done now). More explicit, but in order to decouple the text labels later, we also need the `metadata/text_length` labels which can be annoying for the user (especially since the text lengths should not be intuitively needed)
2. Padding the tensors with -1s so they can be stacked and later we just remove the padding. Could be more confusing on the implementation side, but that should be hidden from regular users. Would not require the extra `text_length` label.

I would personally go with option 2, but do y'all have any thoughts on this?

klemen1999: I think 2 is fine as long as we document that this is automatically happening.

sokovninn: Regarding 2, where should padding and unpadding be done so it remains hidden from the user?

kozlov721: I would just add it to the `prepare` methods. Or more specifically to the `prepare_batch_targets` helper function in OCR utils.

kozlov721: So instead of `targets_batch` and `target_lengths` it would just take the `targets_batch`.

sokovninn: I've added padding in the collate function with zeros, which simplified preprocessing a lot. In Unicode the null character ('\x00') is mapped to 0, so it shouldn't cause any issue.

luxonis_train/config/predefined_models/ocr_recognition_model.py
```diff
@@ -55,7 +55,7 @@ def __init__(
         loss_params: Params | None = None,
         visualizer_params: Params | None = None,
         task_name: str | None = None,
-        alphabet: list[str] | None = None,
+        alphabet: list[str] | str = "english",
```
klemen1999, line 58: Maybe have it as a `Literal["english", ....]` instead of `str` so it's typechecked for all actually possible strings.

luxonis_train/nodes/README.md
```diff
@@ -319,13 +319,13 @@ Adapted from [here](https://arxiv.org/abs/2108.07610).
...
-| Key              | Type    | Default value | Description                           |
-| ---------------- | ------- | ------------- | ------------------------------------- |
-| `alphabet`       | `list`  | `None`        | List of characters.                   |
-| `ignore_unknown` | `bool`  | `True`        | Whether to ignore unknown characters. |
-| `fc_decay`       | `float` | `0.0004`      | L2 regularization factor.             |
-| `mid_channels`   | `int`   | `None`        | Number of middle channels.            |
-| `return_feats`   | `bool`  | `False`       | Whether to return features.           |
+| Key              | Type               | Default value | Description                                            |
+| ---------------- | ------------------ | ------------- | ------------------------------------------------------ |
+| `alphabet`       | `list[str] \| str` | `None`        | List of characters or a name of a predefined alphabet. |
```
klemen1999, line 324: Let's change to Literal and mention all available alphabets (same for README in predefined_models).

klemen1999 (approved):
Generally LGTM, just couple of small comments

luxonis_train/config/predefined_models/ocr_recognition_model.py
sokovninn: Wouldn't such a Literal be too long? Also, it could be expanded frequently, which would make it difficult to maintain.

klemen1999: IMO it just not "looking nice" shouldn't be that big of a point against it where on the other side we gain better type checking. We can also create a new type e.g. `AlphabetName = Literal["english",...]` on top and then `alphabet: list[str] | AlphabetName`? 
On the expandabiliy: I don't think this should be an issue, right now we have a lot already covered and if we add more we just expand this list of literals, shouldn't be that much of an issue IMO.

klemen1999 (approved):

PR #161, opened 2025-01-27
fix assigners

JSabadin:
This PR adjusts the assigners to function in the same manner.

### DIFF
--- a/luxonis_train/assigners/tal_assigner.py
+++ b/luxonis_train/assigners/tal_assigner.py
@@ -254,4 +254,9 @@ class TaskAlignedAssigner(nn.Module):
             torch.full_like(assigned_scores, 0),
         )
 
+        assigned_labels = torch.where(
+            mask_pos_sum.bool(),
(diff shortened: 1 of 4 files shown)


kozlov721 (approved):

luxonis_train/assigners/tal_assigner.py
```diff
@@ -254,4 +254,9 @@ def _get_final_assignments(
             torch.full_like(assigned_scores, 0),
         )
 
+        assigned_labels = torch.where(
```
klemen1999, line 257: Doesn't ATTS assigner also need this?

klemen1999 (approved):

luxonis_train/assigners/tal_assigner.py
JSabadin: It already has it. I removed that line in the instance segmentation commit, and it’s better to have that line inside the assigner rather than in the loss function.

PR #163, opened 2025-01-29
Fix Keypoint & Instance Segmentation Training Hyperparameters 

JSabadin:
# 🔥 Updated Hyperparameters  

## What's New?  
- Added optimized hyperparameters for **keypoint detection** and **instance segmentation**  
- Experimentally determined for best performance  .
- Tested on **Tiger** and **Crack** datasets, achieving **SOTA performance** 🚀  

## Results  
image  
image

### DIFF
--- a/configs/instance_segmentation_light_model.yaml
+++ b/configs/instance_segmentation_light_model.yaml
@@ -16,8 +16,9 @@ loader:
     dataset_name: coco_test
 
 trainer:
+  precision: "16-mixed"
   preprocessing:
-    train_image_size: [384, 512]
+    train_image_size: [384, 384]
     keep_aspect_ratio: true
     normalize:
       active: true
@@ -27,10 +28,10 @@ trainer:
 
   batch_size: 8
   epochs: &epochs 300
-  accumulate_grad_batches: 8 # For best results, always accumulate gradients to effectively use 64 batch size
-  n_workers: 8
+  n_workers: 4
   validation_interval: 10
   n_log_images: 8
+  gradient_clip_val: 10
 
   callbacks:
     - name: EMACallback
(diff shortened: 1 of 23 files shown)


configs/instance_segmentation_light_model.yaml
```diff
@@ -16,8 +16,9 @@ loader:
     dataset_name: coco_test
 
 trainer:
+  precision: "16-mixed"
   preprocessing:
-    train_image_size: [384, 512]
+    train_image_size: [384, 384]
```
klemen1999, line 21: Is this intentional?

luxonis_train/core/utils/train_utils.py
```diff
@@ -19,7 +19,6 @@ def create_trainer(cfg: TrainerConfig, **kwargs: Any) -> pl.Trainer:
         devices=cfg.devices,
         strategy=cfg.strategy,
         max_epochs=cfg.epochs,
-        accumulate_grad_batches=cfg.accumulate_grad_batches,
```
klemen1999, line 22: The reason for removing it here is that we manually handle it in `configure_callbacks()` so we don't apply it twice in case `GradientAcummulationScheduler` is also used?

klemen1999 (commented):
Generally LGTM, just left some comments

configs/instance_segmentation_light_model.yaml
JSabadin: No, it is not intentional. Will fix it.

luxonis_train/core/utils/train_utils.py
JSabadin: Yeah, exactly. These two can't work together. PL will raise an error. Additionaly, PL always sets accumulate_grad_batches to 1 by default

luxonis_train/config/config.py
```diff
@@ -201,6 +201,32 @@ def check_graph(self) -> Self:
...
+            self.nodes,
+            self.losses,
+            self.metrics,
+            self.visualizers,
+        ]:
+            invalid_char_error_message = (
+                "Name, alias or attached module contains a '/', which is not allowed. "
+                "Please rename the node or module to remove any '/' characters."
+            )
+            for module in modules:
```
kozlov721, line 216: I think this is not entirely correct. The attached module checks should also check `alias` instead of just `name`. And I think it is enough in both cases to check just the name/alias.

```python
for module in modeles:
    if "/" in (module.alias or module.name):
        raise ...
```

`attached_to` must reference an existing alias so if the name there would contain slashes it will fail elsewhere that `attached_to` references a node that does not exist in the network.

luxonis_train/config/config.py
```diff
@@ -201,6 +201,32 @@ def check_graph(self) -> Self:
...
 
+    @model_validator(mode="after")
+    def check_for_invalid_characters(self) -> Self:
+        for modules in [
+            self.nodes,
+            self.losses,
+            self.metrics,
+            self.visualizers,
+        ]:
+            invalid_char_error_message = (
```
kozlov721, line 212: We should also include the wrong name in the error message.

kozlov721 (approved):
LGTM, just left a small comment

PR #167, opened 2025-02-04
Fix Bug in Anomaly Detection Model

JSabadin:
## Purpose

- Fixed a bug in the anomaly detection head where an extra normalization was applied. Additionally, returning zero masks is beneficial, as users might mistakenly set non-zero masks for the training dataset, leading to incorrect cases where Perlin noise is not applied.
- Fixed the issue where predefined N model used to have a heavy head instead of a nano one. Now, its RVC2 speed is 10 FPS.
- Mixed-precision training is now supported (stable training) and can significantly speed up training.

| Precision       | Image Size | Dataset Size | Batch Size | Hardware        | Epoch Time |
|----------------|-------------|------------|---------------|-----------------|--------------|
| FP16 (Mixed) | 256×256    |  1.1k         | 16                 | g2standard16  | ~50s           |
| FP32 (Full)  | 256×256       |  1.1k          | 16                | g2standard16  | ~90s           |

## Specification

Training dataset should be created like:
```python
# Training: yield only the 'normal image' class
def train_generator(train_paths):
        ...
        yield {
            "file": path,
            "annotation": {
                "class": "Anomaly",
            },
        }

# Validation/Testing: yield 'anomaly image' with segmentation
def val_test_generator(val_test_paths):
        ...
        yield {
            "file": path,
            "annotation": {
                "class": "Anomaly",
                "segmentation": {
                    "height": image_height,
                    "width": image_width,
                    "points": poly_normalized,
                },
            },
        }
```

- No new dependencies introduced.
- No breaking changes are expected.

## Testing & Validation

- Verified that training runs produce correct masks and image classes.
- Observed faster speed using mixed precision. 
- Trained on the VIS dataset (PCB class) with 1.1k images using augmentations (rotation, flip, affine, perspective) and achieved the following results after 200 epochs (although ideally, training should be run for 700 epochs):
<p align="left">
  
  
</p>
Note that the plot had the background and cracks swapped; this issue was fixed in this PR so that it works correctly.

### DIFF
--- a/luxonis_train/config/predefined_models/anomaly_detection_model.py
+++ b/luxonis_train/config/predefined_models/anomaly_detection_model.py
@@ -33,7 +33,7 @@ def get_variant(variant: VariantLiteral) -> AnomalyVariant:
         "heavy": AnomalyVariant(
             backbone="RecSubNet",
             backbone_params={"variant": "l"},
-            head_params={"variant": "n"},
+            head_params={"variant": "l"},
         ),
     }
 
--- a/luxonis_train/config/predefined_models/anomaly_detection_model.py
+++ b/luxonis_train/config/predefined_models/anomaly_detection_model.py
@@ -106,7 +106,7 @@ class AnomalyDetectionModel(BasePredefinedModel):
             MetricModuleConfig(
                 name="JaccardIndex",
                 attached_to=f"{self.task_name}-DiscSubNetHead",
-                params={"num_classes": 1, "task": "binary"},
+                params={"num_classes": 2, "task": "multiclass"},
                 is_main_metric=True,
             ),
         ]
--- a/luxonis_train/config/predefined_models/anomaly_detection_model.py
+++ b/luxonis_train/config/predefined_models/anomaly_detection_model.py
@@ -18,6 +18,7 @@ VariantLiteral: TypeAlias = Literal["light", "heavy"]
 class AnomalyVariant(BaseModel):
     backbone: str
     backbone_params: Params
+    head_params: Params
 
 
 def get_variant(variant: VariantLiteral) -> AnomalyVariant:
@@ -27,10 +28,12 @@ def get_variant(variant: VariantLiteral) -> AnomalyVariant:
         "light": AnomalyVariant(
             backbone="RecSubNet",
             backbone_params={"variant": "n"},
+            head_params={"variant": "n"},
         ),
         "heavy": AnomalyVariant(
             backbone="RecSubNet",
             backbone_params={"variant": "l"},
+            head_params={"variant": "n"},
         ),
     }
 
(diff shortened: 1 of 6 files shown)


luxonis_train/loaders/base_loader.py
```diff
@@ -275,18 +275,18 @@ def dict_numpy_to_torch(
 
         return torch_dictionary
 
-    def read_image(self, path: str) -> npt.NDArray[np.float32]:
+    def read_image(self, path: str) -> npt.NDArray[np.uint8]:
```
klemen1999, line 278: Let's specify in the docstring that it returns an unnormalized image

klemen1999 (approved):
Generally LGTM

> but training with mixed precision is unstable (loss is not converging)

Did we try to use gradient clipping for that case?

luxonis_train/config/predefined_models/anomaly_detection_model.py
```diff
@@ -103,7 +106,7 @@ def metrics(self) -> list[MetricModuleConfig]:
             MetricModuleConfig(
                 name="JaccardIndex",
                 attached_to=f"{self.task_name}-DiscSubNetHead",
-                params={"num_classes": 2, "task": "multiclass"},
+                params={"num_classes": 1, "task": "binary"},
```
kozlov721, line 109: I think this should still be `"multiclass"`. It is technically binary, but we're explicitly using 2 classes for background and anomaly (line 101 in `perlin_loader.py`)

klemen1999 (approved):

luxonis_train/loaders/luxonis_perlin_loader_torch.py
```diff
@@ -71,27 +71,32 @@ def get(self, idx: int) -> tuple[Tensor, Labels]:
...
+                    tensor_img, anomaly_img, self.beta
+                )
+            else:
+                aug_tensor_img = tensor_img
+                an_mask = torch.zeros((self.height, self.width))
         else:
             aug_tensor_img = tensor_img
+            an_mask = torch.tensor(
+                labels.pop(f"{self.task_name}/segmentation")
+            )[-1, ...]
```
kozlov721, line 99: So, are we expecting simple binary mask (1, H, W) or explicit 2-class mask (2, H, W)? 
If the former, then this should be changed to `squeeze`. Or better we should check first if the channel dimension is 1 and fail with an explanatory error message (and this should be done in the case when we expect explicit 2 classes as well). 
IMHO making it all binary (in the head and loss as well) would be the simplest, but I don't know if we might want to support anomaly detection with multiple classes for different kinds of anomalies.

klemen1999: > IMHO making it all binary (in the head and loss as well) would be the simplest, but I don't know if we might want to support anomaly detection with multiple classes for different kinds of anomalies.

I think for now support it just for binary is fine. If there is a need for multiple classes we can introduce this feature in future releases

PR #175, opened 2025-02-12
Sort classes alphabetically

JSabadin:
## Purpose

This PR is connected to the loader PR provided [here](https://github.com/luxonis/luxonis-ml/pull/245).


luxonis_train/loaders/luxonis_loader_torch.py
```diff
@@ -119,7 +119,12 @@ def get(self, idx: int) -> tuple[Tensor, Labels]:
 
     @override
     def get_classes(self) -> dict[str, list[str]]:
```
kozlov721, line 121: We should change the return type to the class mapping instead of extracting just the names so we can have more flexibility

PR #177, opened 2025-02-14
feat: add per-class weighting option to AdaptiveDetectionLoss

jkbmrz:
## Purpose
Currently, when training an object detection model, we put equal emphasis on each object class. However, if object classes in dataset are not balanced, the training might benefit from putting more emphasis on the under-represented classes. This feature is introduced here by adding the per-class-weighting to the relevant loss function.

## Specification
The `AdaptiveDetectionLoss` class is getting an optional parameter `per_class_weights`. It accepts a list of float values representing the weights by which the losses related to each object class are multiplied.

The new feature is optional so it introduces no dependency disruptions.

## Deployment Plan
/

## Testing & Validation
An over-fitting dataset was prepared including 10 copies of a single image (with minor pixel modification) with 3 annotated objects of different classes. The data was split 8:1:1 among the train, val, and test subsets. 

First, we made sure that not setting the `per_class_weights` parameter (or setting all class weights to 1.0) returns the same results as if using the old version of the loss function. We also made sure in this training run that the model is capable of learning all of the 3 classes (test MAP=1). 

Second, we tried over-fitting for a single class by setting one of the class weights to 1.0 and others to 0.0. As expected, this resulted in models capable of recognizing only objects of a single class they were trained for (test MAP=0.333).


luxonis_train/attached_modules/losses/adaptive_detection_loss.py
```diff
@@ -264,12 +268,18 @@ def __init__(self, alpha: float = 0.75, gamma: float = 2.0):
         self.gamma = gamma
 
     def forward(
-        self, pred_score: Tensor, target_score: Tensor, label: Tensor
+        self,
+        pred_score: Tensor,
+        target_score: Tensor,
+        label: Tensor,
+        per_class_weights: Tensor,
```
klemen1999, line 275: I would make this part of the Varifocal init (class attribute) since it is static, doesn't change during runtime. And also make it into torch tensor there so `torch.tensor(self.per_class_weights)` is not called on every forward (refering to code below)

luxonis_train/attached_modules/losses/adaptive_detection_loss.py
```diff
@@ -42,6 +42,7 @@ def __init__(
         reduction: Literal["sum", "mean"] = "mean",
         class_loss_weight: float = 1.0,
         iou_loss_weight: float = 2.5,
+        per_class_weights: Optional[List[float]] = None,
```
klemen1999, line 45: Luxonis-train is py3.10 so type annotation can be:
```suggestion
        per_class_weights: list[float] | None = None,
```

luxonis_train/attached_modules/losses/adaptive_detection_loss.py
```diff
@@ -59,6 +60,8 @@ def __init__(
         @param class_loss_weight: Weight of classification loss.
         @type iou_loss_weight: float
         @param iou_loss_weight: Weight of IoU loss.
+        @type per_class_weights: Optional[Tensor]
+        @param per_class_weights: Holds a weight for each class to calculate the weighted loss.
```
klemen1999, line 64: Let's explicitly say that this should match number of classes

luxonis_train/attached_modules/losses/adaptive_detection_loss.py
```diff
@@ -78,6 +81,7 @@ def __init__(
         self.varifocal_loss = VarifocalLoss()
         self.class_loss_weight = class_loss_weight
         self.iou_loss_weight = iou_loss_weight
+        self.per_class_weights = per_class_weights
```
klemen1999, line 84: We should make sure that length of this list is same as `self.n_classes`. If it isn't then IMO we can log a warning and set self.per_class_weights value to None

luxonis_train/attached_modules/losses/adaptive_detection_loss.py
klemen1999: Or even better, call torch.tensor(per_class_weights) already in the AdaptiveDetectionLoss init at the start and have Varifocal loss init then take `...., per_class_weights: Tensor | None = None):`

klemen1999 (commented):
When nailed down new parameter should also be noted in the README [here](https://github.com/luxonis/luxonis-train/tree/main/luxonis_train/attached_modules/losses)

luxonis_train/attached_modules/losses/adaptive_detection_loss.py
jkbmrz: Added in [e290a5b](https://github.com/luxonis/luxonis-train/pull/177/commits/e290a5b0a10988e1586a328c85ee29b457a8b507).

luxonis_train/attached_modules/losses/adaptive_detection_loss.py
jkbmrz: Added in [e290a5b](https://github.com/luxonis/luxonis-train/pull/177/commits/e290a5b0a10988e1586a328c85ee29b457a8b507). However, I've set it to error out if there is a mismatch.

luxonis_train/attached_modules/losses/adaptive_detection_loss.py
```diff
@@ -75,7 +78,17 @@ def __init__(
...
         )
 
-        self.varifocal_loss = VarifocalLoss()
+        if len(per_class_weights) != self.n_classes:
+            raise ValueError(
+                f"Incorrect per_class_weights length. Expected {self.n_classes} but got {len(per_class_weights)}."
+            )
+        self.per_class_weights = torch.tensor(
+            per_class_weights
+        )  # TODO: set device?
```
klemen1999, line 87: Could you test if it works with CPU and CUDA training (you can set `accelerator` explicitly to change between them)? Afaik PL should take care of putting tensor on correct device but let's double check

luxonis_train/attached_modules/losses/adaptive_detection_loss_weighted.py
klemen1999, line 1: Probably shouldn't be part of the PR?

klemen1999 (commented):
Generally LGTM, just left small commnets

luxonis_train/attached_modules/losses/adaptive_detection_loss_weighted.py
jkbmrz: Accidentally commited. Removed in [74c4bf0](https://github.com/luxonis/luxonis-train/pull/177/commits/74c4bf0dfe86adbbcab5970f03ff869f2db53df5).

luxonis_train/attached_modules/losses/adaptive_detection_loss.py
jkbmrz: Added in [63a4163](https://github.com/luxonis/luxonis-train/pull/177/commits/63a4163b85fc2c61fa50c2e44cd7d6c8e0a082a4).

klemen1999 (approved):
LGTM

PR #179, opened 2025-02-18
Fix keypoints loss and assignment when no annotations are present

JSabadin:
## Purpose

Fix keypoints loss and assignment when no annotations are present.


kozlov721 (approved):
LGTM. It seems to be working, thanks!

PR #181, opened 2025-02-18
Removed `semgrep` from pre-commit hooks

kozlov721:
## Purpose

`semgrep` is not supported on Windows, so having it in pre-commit hooks blocks development for Windows users. We already have a specialized workflow for `semgrep` so the pre-commit hook is redundant. 

## Specification

Removed `semgrep` hook from `pre-commit-config.yaml`


JSabadin (approved):
LGTM!

PR #190, opened 2025-03-13
Fix the prototype's instance segmentation model attach index

JSabadin:
## Purpose

Fix the prototype's instance segmentation model attach index. Updated the README with FPS for RVC2 and RVC4.

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


klemen1999 (approved):
if verified that it is still training as expected then LGTM

PR #192, opened 2025-03-14
fix missing task names

JSabadin:
## Purpose

This is a fix for task names. Previously, we did not log correctly when the dataset had a task, such as `coco`. For example, we logged `val/visualizations/-EfficientBBoxHead/BBoxVisualizer` instead of `val/visualizations/coco-EfficientBBoxHead/BBoxVisualizer,` and the same issue occurred for loss and metrics.

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/luxonis_train/models/luxonis_lightning.py
+++ b/luxonis_train/models/luxonis_lightning.py
@@ -188,6 +188,14 @@ class LuxonisLightningModule(pl.LightningModule):
             task_names = list(self.dataset_metadata.task_names)
             if not node_cfg.task_name:
                 if len(task_names) == 1:
+                    if node_cfg.task_name != task_names[0] and issubclass(
+                        Node, BaseHead
+                    ):
+                        raise ValueError(
+                            f"Dataset contains a single task: {task_names}. "
+                            f"Node {node_name} does not have the `task_name` parameter set. "
+                            "Please specify the `task_name` parameter for each head node."
+                        )
                     node_cfg.task_name = task_names[0]
                 elif issubclass(Node, BaseHead):
                     raise ValueError(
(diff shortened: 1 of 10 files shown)


luxonis_train/models/luxonis_lightning.py
```diff
@@ -502,6 +510,17 @@ def forward(
...
+    def _format_node_name(self, node_name: str) -> str:
+        """Format node name with task name.
+
+        @type node_name: str
+        @param node_name: Original node name
+        @rtype: str
+        @return: Formatted node name with task name prefix if available
+        """
+        task_name = self.node_task_names[node_name]
+        return f"{task_name}-{node_name}"
```
kozlov721, line 522: Would this PR make it possible to replace `-` with `/` so it is more consistent, and IMO looks better (especially when the `task_name` is an empty string). I know it caused some issues before but since it's no longer part of the actual node alias it could be okay now?

kozlov721 (approved):
Left one suggestion open for further discussion, but more of a nice-to-have. Otherwise LGTM

luxonis_train/models/luxonis_lightning.py
JSabadin: If we replaced it there, the logging would fail with the following error:
`Failed to re-log stored logs to MLflow: INVALID_PARAMETER_VALUE: Invalid metric name: 'train/loss//EfficientBBoxHead/AdaptiveDetectionLoss'. Names may be treated as files in certain cases and must not resolve to other names when treated as such. This name would resolve to 'train/loss/EfficientBBoxHead/AdaptiveDetectionLoss'.`

JSabadin: @kozlov721 Can we stick to what HubAI needs for now, and we can fix it when a larger task refactor happens?

kozlov721: I suppose the double `/` is problematic. No problem with leaving this for later.

PR #198, opened 2025-03-19
Formatting, linting, general cleanup

kozlov721:
## Purpose

Fixed some type checking errors and added more linter rules to prevent future bugs.

## Specification

- Cleanup of training strategies
- Removed `ModuleFreezer` callback, the module freezing is now handled by the `TrainingManager`
- Added a bunch of linter rules and fixed linting issues in the code

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


pyproject.toml
```diff
@@ -36,16 +36,82 @@ where = ["."]
...
 
 [tool.ruff]
 target-version = "py310"
 line-length = 79
 indent-width = 4
 
 [tool.ruff.lint]
-ignore = ["F403", "B028", "B905", "D1", "W191"]
-select = ["E4", "E7", "E9", "F", "W", "B", "I"]
+select = [
```
kozlov721, line 47: Not all of the rules are enabled for now. The rest will be enabled with the nodes cleanup (which should be the last cleanup PR)

luxonis_train/callbacks/metadata_logger.py
```diff
@@ -45,9 +44,11 @@ def on_fit_start(
             hparams["luxonis_train"] = luxonis_train_hash
 
         pl_module.logger.log_hyperparams(hparams)
-        with open(osp.join(pl_module.save_dir, "metadata.yaml"), "w") as f:
+        # TODO: Remove
+        with open(pl_module.save_dir / "metadata.yaml", "w") as f:
             yaml.safe_dump(hparams, f, default_flow_style=False)
 
+    # TODO: Is this any useful anymore?
```
klemen1999, line 51: IMO we can remove this

luxonis_train/callbacks/metadata_logger.py
```diff
@@ -45,9 +44,11 @@ def on_fit_start(
             hparams["luxonis_train"] = luxonis_train_hash
 
         pl_module.logger.log_hyperparams(hparams)
-        with open(osp.join(pl_module.save_dir, "metadata.yaml"), "w") as f:
+        # TODO: Remove
```
klemen1999, line 47: Should be able to remove as well since we log it with tracker

luxonis_train/core/core.py
```diff
@@ -793,6 +813,7 @@ def _mult(lst: list[float | int]) -> list[float]:
 
         return Path(archive_path)
 
+    # TODO: Where are these methods used? Can they be removed?
```
klemen1999, line 816: These were used early on starting runs in a different thread and then getting status of the run from it. IMO we cna remove all of this now together with ability to start runs in a separate thread as I don't currently see a usecase for it anymore.

luxonis_train/nodes/heads/base_head.py
```diff
@@ -1,10 +1,14 @@
...
+
 from luxonis_train.nodes.base_node import (
     BaseNode,
     ForwardInputT,
     ForwardOutputT,
 )
 
 
+# TODO: We shouldn't skip the head completely
+# if custom config is not defined.
```
klemen1999, line 11: Could you expand a bit on this comment?

klemen1999 (commented):
Generally LGTM, left some comments

luxonis_train/nodes/heads/base_head.py
kozlov721: Currently if the head does not specify the custom config, we skip it during archiving altogether (see [here](https://github.com/luxonis/luxonis-train/blob/178b5eb989ae73cdb9894ba7a216b712e6d50ea9/luxonis_train/core/utils/archive_utils.py#L145)). But we should still include the outputs even if the custom config is not implemented. (Unless this doesn't make sense in the NN Archive?)

klemen1999: I see. Yeah I think we should still include it even if no custom config is specified. In that case the information about it in NNarchive would be minimal but it should still be there imo

klemen1999 (approved):

PR #199, opened 2025-03-19
Feature: FOMO Confusion Matrix

JSabadin:
## Purpose

When logging the confusion matrix (CM), we now also log the Matthews correlation coefficient (MCC). This is because we can use it as the main metric.  

FOMO should have a single metric to determine when to save the weights—now, we have one. OKS lacks a penalty for false positives (FP). Due to the nature of the problem, where finding the object's center is challenging, OKS initially produces very high values because the model predicts a lot of keypoints per image in the early stages. As training progresses, the OKS decreases and then starts to increase again... This is an issue with the model itself. Now, the CMM monotonically increases, so we are actually saving the best model during training.

Additionally, the focal loss term has been added to the FOMO loss.  

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


luxonis_train/attached_modules/metrics/confusion_matrix/utils.py
```diff
@@ -37,6 +37,33 @@ def preprocess_instance_masks(
...
+    sum_diag = torch.diag(cm).sum()
+    sum_rows = cm.sum(dim=1)
+    sum_cols = cm.sum(dim=0)
+    numerator = sum_diag * N - torch.dot(sum_rows, sum_cols)
+    denominator = torch.sqrt(
+        (N**2 - (sum_rows**2).sum()) * (N**2 - (sum_cols**2).sum())
+    )
+
+    if denominator == 0:
+        return torch.tensor(0.0)
```
klemen1999, line 62: Not 100% sure but do we need to specify device when we create this tensor or will this be handled automatically to match the numerator device?

luxonis_train/attached_modules/metrics/confusion_matrix/instance_segmentation_confusion_matrix.py
```diff
@@ -37,7 +37,13 @@ def update(
 
     @override
     def compute(self) -> dict[str, Tensor]:
+        det_cm = DetectionConfusionMatrix.compute(self)
+        rec_cm = RecognitionConfusionMatrix.compute(self)
         return {
-            "detection": DetectionConfusionMatrix.compute(self),
-            "segmentation": RecognitionConfusionMatrix.compute(self),
+            "detection_mcc": compute_mcc(det_cm["confusion_matrix"].float()),
```
klemen1999, line 43: This one can't be used as main metric because we would search for "mcc" in the dictionary which this one wouldn't have right?

luxonis_train/attached_modules/metrics/confusion_matrix/fomo_confusion_matrix.py
```diff
@@ -0,0 +1,111 @@
...
+    ) -> None:
+        """Override update to convert FOMO keypoints into bounding boxes
+        before calling the parent update method."""
+        pred_bboxes = self._keypoints_to_bboxes(
+            keypoints,
+            self.original_in_shape[1],
+            self.original_in_shape[2],
+        )
+
+        super().update(pred_bboxes, target_boundingbox)
```
klemen1999, line 39: Here a TP is if a smaller 5x5 box is predicted inside the target bbox. This would mean that metric would be also high if the model predicted these centers in e.g. upper corner of the target bbox. Maybe it would make more sence to also adjust the target_boundingbox to be same size (5x5) around the center of the bbox so that TP would only be if the prediction overlaps with the smaller bbox in the center of the object we are trying to detect. Thoughts?

luxonis_train/nodes/heads/fomo_head.py
```diff
@@ -17,7 +17,7 @@ def __init__(
         self,
         num_conv_layers: int = 3,
         conv_channels: int = 16,
-        use_nms: bool = False,
+        use_nms: bool = True,
```
klemen1999, line 20: We should update a bit the docs [here](https://github.com/luxonis/luxonis-train/tree/fix/fomo-cm-metric/luxonis_train/config/predefined_models#fomomodel) regarding this. Also it seems like one paragraph is copied over twice

klemen1999 (commented):
Generally LGTM, left some comments

luxonis_train/attached_modules/metrics/confusion_matrix/instance_segmentation_confusion_matrix.py
JSabadin: This will work from now on. When we detect the ConfusionMatrix, we rename its main metric to MCC. See luxonis_lighting.py.

luxonis_train/attached_modules/metrics/confusion_matrix/fomo_confusion_matrix.py
JSabadin: Hmm, if you ask me, this nuance isn't necessary since the center of an object can be anywhere within the bounding box—even near the corner. When testing this on a car dataset from Roboflow, this approach seemed to work best.

luxonis_train/attached_modules/metrics/confusion_matrix/fomo_confusion_matrix.py
```diff
@@ -0,0 +1,111 @@
...
+
+    @override
+    def compute(self) -> Tensor:
+        return {
+            "mcc": compute_mcc(self.confusion_matrix.float()),
+            "confusion_matrix": self.confusion_matrix,
+        }
+
+    @staticmethod
+    def _keypoints_to_bboxes(
```
kozlov721, line 49: We can move this method to the `utils.boundingbox` module.

luxonis_train/attached_modules/metrics/confusion_matrix/instance_segmentation_confusion_matrix.py
```diff
@@ -37,7 +37,15 @@ def update(
...
+        det_renamed = {
+            "detection_mcc": det_result["mcc"],
+            "detection_confusion_matrix": det_result["confusion_matrix"],
         }
+        rec_renamed = {
+            "segmentation_mcc": rec_result["mcc"],
+            "segmentation_confusion_matrix": rec_result["confusion_matrix"],
+        }
+
+        return {**det_renamed, **rec_renamed}
```
kozlov721, line 51: Just a nitpick, but in `3.10` one can use the `|` operator for merging dictionaries ( `return det_renamed | rec_renamed`)

luxonis_train/attached_modules/metrics/confusion_matrix/fomo_confusion_matrix.py
```diff
@@ -0,0 +1,111 @@
...
+        pred_bboxes = self._keypoints_to_bboxes(
+            keypoints,
+            self.original_in_shape[1],
+            self.original_in_shape[2],
+        )
+
+        super().update(pred_bboxes, target_boundingbox)
+
+    @override
+    def compute(self) -> Tensor:
```
kozlov721, line 42: The `compute` method doesn't need to be overridden (it does the same thing as `DetectionConfusionMatrix.compute`

kozlov721 (approved):
LGTM, just left a few small comments

PR #208, opened 2025-03-28
Fix: Resuming Training from Checkpoint

JSabadin:
## Purpose

Now, the weights passed to `model.train()` will function the same way as those in the config. So, to resume training, you need to set the `resume_training `flag in the `config.yaml` file.

## Specification

None / not applicable

None / not applicable

IMPORTANT: Python API will change from `.train(resume_weights:...)` to `.train(weights=...)` to be more inline with what it actually does.

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/configs/detection_light_model.yaml
+++ b/configs/detection_light_model.yaml
@@ -2,6 +2,7 @@
 # NOTE: This example downloads pretrained COCO weights and training parameters are already prepared for fine tuning
 
 model:
+  weights:  "C:\\Users\\jerne\\Desktop\\luxonis-train-ml\\luxonis-train\\output\\197-salmon-swan\\best_val_metric\\detection_light_EfficientBBoxHead_MeanAveragePrecision=nan_loss=24.4528_09.ckpt"
   name: detection_light
   predefined_model:
     name: DetectionModel
@@ -25,6 +26,7 @@ loader:
 
 trainer:
   precision: "16-mixed"
+  # resume_training: true
   preprocessing:
     train_image_size: [384, 512]
     keep_aspect_ratio: true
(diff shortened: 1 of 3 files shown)


luxonis_train/core/core.py
```diff
@@ -254,10 +255,20 @@ def train(
...
+            if self.cfg.model.weights is not None:
+                logger.warning(
+                    "Weights provided in the command line, but config weights are set. "
+                    "Ignoring weights provided in config."
+                )
+            self.lightning_module.load_checkpoint(weights)
+        else:
+            weights = self.cfg.model.weights
+
+        resume_weights = weights if self.cfg.trainer.resume_training else None
```
klemen1999, line 271: Let's add another warning if resume_training=True and resume_weights after this line is None that the resume_training is specified but no weights to resume from so the trainnig will start withouth any loaded weights

configs/detection_light_model.yaml
klemen1999, line 1: Probably shouldn't be included in the PR?

luxonis_train/__main__.py
```diff
@@ -39,18 +39,18 @@ def train(
...
 ):
     """Start the training process.
 
     @type config: str
     @param config: Path to the configuration file.
-    @type resume_weights: str
-    @param resume_weights: Path to the model weights to resume training
-        from. @type *opts: tuple[str, str]
+    @type weights: str
+    @param weights: Path to the model weights. from. @type *opts:
```
klemen1999, line 49: Should probably be something like:
    @param weights: Path to the model weights.
    @type *opts: tuple[str,str]

klemen1999 (approved):
LGTM, left some comments

PR #216, opened 2025-04-02
Fix: loss, archive, ema

JSabadin:
## Purpose

This PR fixes the following issues:

- **Archive Command on Windows:**
  - Resolved issues with the archive command failing on Windows.

- **Accumulator Loss Reset:**
  - Fixed a bug where the accumulator was not being reset properly.
  - Previously, the train accumulator was used for validation and testing, leading to incorrect loss values being logged, appearing identical to the train values.

- **`current_epoch` for Nodes:**
  - Corrected an issue where `current_epoch` was always set to `0` for nodes.

- **Bug in EMA (Exponential Moving Average):**
  - Fixed improper loading of weights in EMA.
  - This issue caused incorrect training behavior and resulted in slower training performance.

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

This PR fixes the EMA tests and adds a unit test for the training strategy.

### DIFF
--- a/luxonis_train/callbacks/ema.py
+++ b/luxonis_train/callbacks/ema.py
@@ -245,18 +245,6 @@ class EMACallback(pl.Callback):
         """
         self._swap_to_ema_weights(pl_module)
 
-    def on_train_epoch_start(
-        self, trainer: pl.Trainer, pl_module: pl.LightningModule
-    ) -> None:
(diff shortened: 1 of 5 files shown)


klemen1999 (approved):
Checked locally against commit before corruption and we get same results so LGTM

PR #221, opened 2025-04-15
Fix/tuner issues

JSabadin:
### Tuner Improvements
- Option to monitor either the main validation metric or validation loss.
- Save `tuner_study.csv` in the run directory.
- Fixed the issue where TensorBoard did not display nicely nested logging runs.  
  > **Note:** This requires the related [luxonis-ml PR](https://github.com/luxonis/luxonis-ml/pull/299) to be merged first.
- Added a `get_mlflow_logging_keys` function, which returns the keys logged to MLflow artifacts and metrics — useful for HubAI.

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

Example of tuner trials.  
Checking which trial corresponds to which hyperparameters can be done simply by inspecting the **hparams** tab in TensorBoard.

image


luxonis_train/core/core.py
```diff
@@ -616,8 +619,32 @@ def _objective(trial: optuna.trial.Trial) -> float:
...
+                monitor = "val/loss"
+            elif cfg.tuner.monitor == "metric":
+                main_metric = next(
+                    (m for m in cfg.model.metrics if m.is_main_metric), None
+                )
+                if main_metric:
+                    all_mlflow_logging_keys = self.get_mlflow_logging_keys()
+                    search_name = (
+                        "mcc"
+                        if main_metric.name == "ConfusionMatrix"
```
klemen1999, line 632: Can we override the name property on the ConfusionMatrix class and then access it like we do all others? Main reason is so we don't have to have this if anywhere where ConfMatrix is (also in build_metrics() function). Would this be better long term or would it be maybe breaking something?

luxonis_train/core/core.py
```diff
@@ -857,3 +897,6 @@ def get_best_metric_checkpoint_path(self) -> str | None:
             if callback.monitor and "val/metric/" in callback.monitor:
                 return callback.best_model_path
         return None
+
+    def get_mlflow_logging_keys(self):
```
klemen1999, line 901: Let's add docstrings here as well

luxonis_train/core/core.py
```diff
@@ -687,16 +717,26 @@ def _objective(trial: optuna.trial.Trial) -> float:
...
         study.optimize(
             _objective, n_trials=cfg_tuner.n_trials, timeout=cfg_tuner.timeout
         )
+        logger.info(
+            f"Best study parameters: {study.best_params}. Cost: {study.best_value}."
+        )
 
-        logger.info(f"Best study parameters: {study.best_params}")
+        study_df = study.trials_dataframe()
+        study_df.to_csv(self.run_save_dir / "tuner_study.csv", index=False)
```
klemen1999, line 735: Does this contain all the trials or only the best one? Will it also contain the best study parameters overall?

klemen1999 (commented):
Generally LGTM, just left some questions

luxonis_train/core/core.py
JSabadin: Yeah, I agree.  Should we do this in a separate PR?

luxonis_train/core/core.py
JSabadin: It contain data of all trials. It will also contain information of the hyperparameters for each one of  the trials.

klemen1999 (approved):

PR #226, opened 2025-04-16
Smarter Image Logging

JSabadin:
## Purpose

This PR introduces a smarter image logging strategy. Instead of simply logging the first N images in sequence (which can over-represent common classes), the new approach prioritizes images from under-represented classes. This ensures more balanced monitoring of model performance across all classes.

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/luxonis_train/lightning/luxonis_lightning.py
+++ b/luxonis_train/lightning/luxonis_lightning.py
@@ -117,6 +117,7 @@ class LuxonisLightningModule(pl.LightningModule):
         self._export: bool = False
         self._core = _core
         self._n_logged_images = 0
+        self._class_log_counts: list[int] = []
 
         self._loss_accumulators = {
             "train": LossAccumulator(),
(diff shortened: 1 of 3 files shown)


luxonis_train/lightning/luxonis_lightning.py
```diff
@@ -621,19 +625,43 @@ def _evaluation_step(
...
-                        f"{name}/{self._n_logged_images}",
-                        viz.detach().cpu().numpy().transpose(1, 2, 0),
-                        step=self.current_epoch,
+        max_log_images = self.cfg.trainer.n_log_images
+
+        if outputs.visualizations:
+            if cls_key is not None:
+                # Smart logging: balance class representation
+                n_classes = labels[cls_key].shape[1]
+                if not hasattr(self, "_class_log_counts"):
```
klemen1999, line 634: Would it be cleaner to just have this always as a class attribute (set to None in the init) and then here we just every time set it to [0]*n_classes? And then same below, just set it to list with 0s every time.
I don't think it is an issue if self._class_log_counts is always there (even if smart logging can't be used) as it is private variable anyway and it doesn't pollute the namespace much IMO

luxonis_train/lightning/utils.py
```diff
@@ -531,3 +531,76 @@ def _to_module_dict(modules: AttachedModulesDict[A]) -> AttachedModulesDict[A]:
             for node_name, node_modules in modules.items()
         }
     )  # type: ignore
+
+
+def log_balanced_class_images(
```
klemen1999, line 536: This this still guarantee that we always log the same set of images?

JSabadin: "Yeah, we’re still logging the same set of images evey val epoch, but they’re now distributed more nicely. (We’re not shuffling them, if that’s what you’re asking.)

klemen1999 (approved):

PR #232, opened 2025-05-05
Fix: CLI `--source` argument

kozlov721:
## Purpose

Fixes an option to source custom python modules.
## Specification

Modifies the entry point to the CLI in `pyproject.toml`

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ classifiers = [
 ]
 
 [project.scripts]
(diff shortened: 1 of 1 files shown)


klemen1999 (approved):
LGTM
Can we add some sort of test for the source option? Like simple custom node that we source in the test and then just check if config that reference it can be used to build LuxonisModel?

PR #234, opened 2025-05-07
Add `lr_after_unfreeze`

JSabadin:
### What’s Changed

* **New `lr_after_unfreeze` argument**  
  Allows users to explicitly set the learning rate for parameters that have just been unfrozen.

* **Fallback behavior**  
  If `lr_after_unfreeze` is not provided, the learning rate of the optimizer’s first parameter group is divided by `initial_denom_lr` (in PyTorch-Lightning this previously defaulted to 10; it is now set to 1).

> **Note:**  
> PyTorch Lightning currently creates new parameter groups named `lr-{opt_name}/pg{idx}` after unfreezing (see [source](
> https://github.com/Lightning-AI/pytorch-lightning/blob/d5208f7052252baa3f2c5b59184b6cd38f0c80c4/src/lightning/pytorch/callbacks/finetuning.py#L247-L275)).  
> This naming convention cannot be overridden.

### Why
Previously, Lightning always used `base_lr/10` for any newly added param group when no LR was passed. This update gives you direct control over that LR, and schedulers will apply to the unfrozen group just like any other.

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

### Unfreeze with same LR

<p align="center">
  Unfreeze with same LR
</p>

### Unfreeze with backbone’s new LR

<p align="center">
  Unfreeze with backbone’s new LR
</p>


klemen1999 (commented):
Can we add some tests to this whole `freezing` feature? Like freeze part of the model, check learning rate and if it requrires gradient (should be false when frozen) on this part before and after unfreeze (with and without specifying lr_after_unfreeze)

kozlov721 (approved):

klemen1999 (approved):

PR #243, opened 2025-05-30
Feat: Annotate Directory of Images

JSabadin:
## Purpose

This feature annotates a directory of images by performing predictions with a model and creating a dataset in LDF from those predictions.  

A new command `annotate` is added to the CLI, and you can also use it via `model.annotate(...)`.

The `annotate` functionality uses the config YAML file where the model is defined, and performs inference on the given image directory. For example, if a user defines a detection head like this:

```yaml
nodes:
  ...
  - name: EfficientBBoxHead
    params:
      download_weights: False
      initialize_weights: True
      conf_thres: 0.03
      iou_thres: 0.65
    inputs:
      - RepPANNeck
  ...
```

The `annotate` process will use the `conf_thres` value of `0.03` and `iou_thres` of `0.65` when performing Non-Maximum Suppression (NMS) for that head.

By default, `annotate` supports all built-in `luxonis-train` tasks. If the head operates in a way that's aligned with these default tasks, it will work out of the box since the base head is designed for compatibility.

However, if custom post-processing of labels is needed, the user must implement their own `annotate` function for the head. In such a case, the user can apply any custom filtering logic to the model outputs.

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


tersekmatija (commented):
Do we have some example usage?

Is it possible to configure also "post-processing" parameters like confidence, IoU threshold, ...? I assume that's defined through the architecture itself?

kozlov721 (approved):

PR #253, opened 2025-07-01
Fix Saving Config

JSabadin:
## Purpose

Fixed issue where config saving failed due to `dataset_type` being stored as an enum in the loader config.

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


klemen1999 (commented):
LGTM. Lets expand the tests as well with this (and if you see any other place where we use Enums directly in the config). 
Maybe we can add these tests [here](https://github.com/luxonis/luxonis-train/blob/main/tests/integration/test_model_config.py) or if you have a better idea for their location?

PR #256, opened 2025-07-05
Feature: Debug Mode (Loader)

kozlov721:
## Purpose

Makes it easier to create and test models before the entire pipeline is defined. 

## Specification

- Adds new argument `debug_mode` to `lxt.Core` that sets the model to a debug mode
  - Corresponding CLI flag (`--debug`) added to `test` and `train` commands
- Added new dummy `DebugLoader` that is used instead of `LuxonisLoaderTorch` in case we fail to load the specified dataset (and `debug_mode` is set to `True`)
  - Yields empty images and labels. Compatible with all LDF label types, but can be extended so it can replace custom loaders as well   

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/luxonis_train/core/core.py
+++ b/luxonis_train/core/core.py
@@ -28,8 +28,11 @@ from luxonis_train.callbacks import (
 )
 from luxonis_train.config import Config
 from luxonis_train.lightning import LuxonisLightningModule
-from luxonis_train.loaders import BaseLoaderTorch
-from luxonis_train.loaders.luxonis_loader_torch import LuxonisLoaderTorch
+from luxonis_train.loaders import (
+    BaseLoaderTorch,
+    DummyLoader,
+    LuxonisLoaderTorch,
+)
 from luxonis_train.registry import LOADERS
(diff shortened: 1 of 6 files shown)


luxonis_train/core/core.py
```diff
@@ -143,20 +155,52 @@ def __init__(
...
+                    logger.error(
+                        "Unable to initialize loader. If you want to run "
+                        "the model in debug mode, set `debug_mode=True`."
+                    )
+                    raise
+                logger.warning(
+                    f"Failed to initialize loader '{loader_name}' "
+                    f"for view '{view}'. Using `DummyLoader` instead."
+                )
+                n_keypoints = self.cfg.loader.params.get("n_keypoints", 4)
```
klemen1999, line 184: Any reason why we have 4 as default here but then below we have 3 as default?

kozlov721: Thanks, that's a leftover from debugging it

PR #258, opened 2025-07-11
Custom checkpoint

kozlov721:
## Purpose

Better flexibility in the future regarding backwards compatibility.

## Specification

Adds a few custom keys to the saved checkpoint:
- `"execution_order"` - saves the execution order of modules in a model
- `"config"`: - the configuration 
- `"dataset_metadata"`: Dumped `DatasetMetadata` containing info about number / names of classes, keypoints, etc.

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

Added relevant tests


kozlov721:
> Adds a few custom keys to the saved checkpoint:

Feel free to suggest more useful stuff that could be good to save in the checkpoints.

kozlov721:
That should be miniscule. We're just adding some text when the checkpoint already contains all the weights.

PR #263, opened 2025-07-18
Feature: Disable `rich` logging

kozlov721:
## Purpose

Adds an option to disable rich logging using config's top-level `rich_logging` field.

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/luxonis_train/config/config.py
+++ b/luxonis_train/config/config.py
@@ -14,6 +14,7 @@ from luxonis_ml.utils import (
 )
 from pydantic import (
     Field,
+    ModelWrapValidatorHandler,
     SerializationInfo,
     field_validator,
     model_serializer,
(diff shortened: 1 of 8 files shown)


luxonis_train/config/config.py
```diff
@@ -479,6 +480,16 @@ class TrainerConfig(BaseModelExtraForbid):
 
     training_strategy: ConfigItem | None = None
 
+    @model_validator(mode="before")
+    @classmethod
+    def validate_use_rich_progress_bar(cls, data: Params) -> Params:
```
klemen1999, line 486: Let's mark it deprecated in the config docs and also add this new parameter to the docs

luxonis_train/config/config.py
```diff
@@ -610,6 +622,38 @@ class Config(LuxonisConfig):
 
     ENVIRON: Environ = Field(exclude=True, default_factory=Environ)
 
+    @model_validator(mode="wrap")
+    @classmethod
+    def check_rich_logging(
```
klemen1999, line 628: What is the reason behind this extra function?

klemen1999 (approved):

luxonis_train/config/config.py
kozlov721: To correctly initialize the logger just for the config. We are logging some info during the config validation so we need to call `setup_logging` with the correct argument for `use_rich` before the config validation starts. It's a bit complicated because we have to also deal with the deprecated `trainer.use_rich_progress_bar` field.

PR #267, opened 2025-07-23
Option to download weights from s3 or gcs

kozlov721:
## Purpose

Makes internal testing easier. 

## Specification

Added support for `"s3://"`, `"gs://"` and `"gcs://"` urls by using the `LuxonisFileSystem`.  

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ Get started with `LuxonisTrain` in just a few steps:
    Open the provided URL in your browser to visualize the training progress
 
 > [!NOTE]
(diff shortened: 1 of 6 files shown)


klemen1999 (approved):
LGTM but this only affects the pretrained checkpoints right? For the weights we already support loading from other sources right?

On the failing test: I've seen this one in some other PRs as well. Is it because of some version upgrade to the packages maybe?

PR #270, opened 2025-07-24
Ensure Full Quota of Validation Visualizations Despite Class Imbalance

JSabadin:
## Purpose

This PR ensures we always log the full quota of validation images, even when some classes are missing:

- Return and track `logged_idxs` from `log_balanced_class_images`

- Extract and buffer only the leftover frames not consumed by balanced logging

- Guarantee `n_log_images` total visualizations per epoch by topping up the buffer once per run

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/luxonis_train/lightning/utils.py
+++ b/luxonis_train/lightning/utils.py
@@ -571,7 +571,7 @@ def log_balanced_class_images(
     nodes: Mapping[str, BaseNode],
     visualizations: dict[str, dict[str, Tensor]],
     labels: Labels,
-    cls_key: str,
+    cls_task_keys: list[str],
     class_log_counts: list[int],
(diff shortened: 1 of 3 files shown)


luxonis_train/lightning/utils.py
```diff
@@ -641,6 +647,47 @@ def log_sequential_images(
     return node_logged_images
 
 
+def compute_visualization_buffer(
```
klemen1999, line 650: Let's add some dosctring explanation on what this function is doing

sokovninn (approved):
LGTM! Further improvement might be independent balancing per head/task. We can address it in a future PR.

PR #273, opened 2025-07-29
CI: Re-enabled type checking

kozlov721:
## Purpose

More strict CI checks.

## Specification

- Re-enabled type checking in the CI (with the exception of `luxonis_train.nodes`)
- Enabled a few more linting rules
- Fixed all type / linting issues

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/luxonis_train/config/config.py
+++ b/luxonis_train/config/config.py
@@ -5,7 +5,13 @@ from typing import Annotated, Any, Literal, NamedTuple
 
 from loguru import logger
 from luxonis_ml.enums import DatasetType
-from luxonis_ml.typing import ConfigItem, Params, ParamValue, check_type
+from luxonis_ml.typing import (
+    ConfigItem,
+    Params,
+    ParamValue,
+    PathType,
+    check_type,
+)
 from luxonis_ml.utils import (
     BaseModelExtraForbid,
     Environ,
@@ -21,7 +27,6 @@ from pydantic import (
     model_serializer,
     model_validator,
 )
-from pydantic.main import IncEx
 from pydantic.types import (
     FilePath,
     NonNegativeFloat,
@@ -406,7 +411,7 @@ class PreprocessingConfig(BaseModelExtraForbid):
         return self
 
     @model_serializer
-    def serialize_model(self, info: SerializationInfo):
+    def serialize_model(self, info: SerializationInfo) -> dict[str, Any]:
         data = {
             key: value
             for key, value in self.__dict__.items()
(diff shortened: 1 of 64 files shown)


luxonis_train/utils/general.py
```diff
@@ -96,8 +96,8 @@ def _infer_upscale_factor(in_size: int, orig_size: int) -> int | float:
 
 
 def to_shape_packet(packet: Packet[Tensor]) -> Packet[Size]:
-    """Converts a packet of tensors to a packet of shapes. Used for
-    debugging purposes.
+    """Converts a packet of tensors to a packet of shapes. purposes.
+    Used for. debugging.
```
klemen1999, line 100: typos

luxonis_train/utils/spatial_transforms.py
```diff
@@ -87,7 +85,7 @@ def transform_masks(
     )
     norm_masks = []
     for mask in raw_masks:
-        if keep_aspect_ratio:
+        if ratio is not None:
```
klemen1999, line 88: why change in logic?

kozlov721: It is equivalent to the old condition (if `keep_aspect_ratio` is `True` then `ratio` is not `None`), but in this version `pyright` doesn't complain about `ratio` possibly being `None`.

luxonis_train/config/config.py
```diff
@@ -618,7 +623,7 @@ class Config(LuxonisConfig):
     trainer: TrainerConfig = Field(default_factory=TrainerConfig)
     exporter: ExportConfig = Field(default_factory=ExportConfig)
     archiver: ArchiveConfig = Field(default_factory=ArchiveConfig)
-    tuner: TunerConfig | None = None
+    tuner: TunerConfig = Field(default_factory=TunerConfig)
```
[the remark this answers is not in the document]
kozlov721: I think this was because we had some fields in `TunerConfig` that didn't have any default value, which is no longer the case.

luxonis_train/nodes/heads/base_head.py
```diff
@@ -23,6 +24,7 @@ class BaseHead(BaseNode[ForwardInputT, ForwardOutputT]):
     @ivar parser: Parser to use for the head.
     """
 
+    task: Task
```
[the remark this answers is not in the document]
kozlov721: It shouldn't be. All the heads define `task` anyways and also this should only affect type checking, not runtime.

klemen1999 (approved):

PR #279, opened 2025-08-02
Tests: Reorganization and cleanup

kozlov721:
## Purpose

Makes the tests faster and more robust.

## Specification

- Reorganized the tests 
- Improved coverage
- Removed redundant tests
- Made the CI execution paralellized
  - The tests are split into 4 test suites
    - `unit` - cheap unit tests
    - `predefined` - tests all the predefined models
    - `combinations` - tests all combinations of backbones and heads
    - `misc` - miscelanneus tests, e.g. CLI, tuning, param freezing etc.
    - Individual suites can be run as `pytest -m <suite>`, e.g. `pytest -m predefined` 
  - Each suite runs in a separate runner

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


luxonis_train/core/utils/annotate_utils.py
```diff
@@ -86,6 +88,13 @@ def annotated_dataset_generator(
...
                     head_output, img_paths, model.cfg_preprocessing
-                )
+                ):
+                    if isinstance(record, DatasetRecord):  # pragma: no cover
+                        yield record
+                    else:
+                        # Skips predictions that are invalid,
+                        # e.g. bboxes outside of the clipping range
+                        with suppress(Exception):
+                            yield DatasetRecord(**record)
```
klemen1999, line 100: Could you elaborate a bit what happens here? Comment says we skip the bboxes outside of clipping range, what if part of the bbox is outside but part inside? Is this then clipped and still outputed or do we skip this bbox?

klemen1999 (approved):
LGTM

luxonis_train/core/utils/annotate_utils.py
kozlov721: This means we skip completely invalid boxes (and other annotations). In luxonis-ml we have the automatic clipping check that refuses annotations whose values are outside of [-2, 2]. This test here makes sure we aren't producing an invalid dataset

PR #287, opened 2025-08-19
Raise clear error when using SoftmaxFocalLoss for binary tasks

JSabadin:
## Purpose

Adds a check in `SoftmaxFocalLoss` to raise an error when used for binary tasks, guiding users to use `SigmoidFocalLoss` instead.

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/luxonis_train/attached_modules/losses/softmax_focal_loss.py
+++ b/luxonis_train/attached_modules/losses/softmax_focal_loss.py
@@ -8,7 +8,6 @@ from luxonis_train.attached_modules.losses import BaseLoss
 from luxonis_train.tasks import Tasks
 
 
-# TODO: Add support for multi-class tasks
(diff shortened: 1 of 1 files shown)


luxonis_train/attached_modules/losses/softmax_focal_loss.py
```diff
@@ -48,6 +47,12 @@ def __init__(
             raise ValueError("smooth value should be in [0,1]")
 
     def forward(self, predictions: Tensor, targets: Tensor) -> Tensor:
+        if predictions.size(1) <= 2:
```
kozlov721, line 50: We can check for this in `__init__` already using `self.n_classes`

klemen1999 (approved):

PR #289, opened 2025-08-28
Fix OKS metric in case of batches without annotations 

JSabadin:
## Purpose

In the update loop of OKS, the GT bounding boxes and keypoints skipped empty batches because the` batch_size` parameter wasn’t passed to `instances_from_batch`. As a result, `self.pred_keypoints` had the correct length, but `self.target_keypoints` was shorter, missing entries for those batches.

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/luxonis_train/attached_modules/metrics/object_keypoint_similarity.py
+++ b/luxonis_train/attached_modules/metrics/object_keypoint_similarity.py
@@ -71,11 +71,15 @@ class ObjectKeypointSimilarity(BaseMetric):
         h, w = self.original_in_shape[1:]
         bs = len(keypoints)
 
-        self.pred_keypoints.extend(map(fix_empty_tensor, keypoints))
-
-        for bboxes, kpts in instances_from_batch(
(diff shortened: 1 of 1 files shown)


luxonis_train/attached_modules/metrics/object_keypoint_similarity.py
klemen1999, line 1: Do we have some test that would catch possible regression here? Or did you perform some manual testing with actual values to make sure the end metric is correct?

klemen1999: And now if image has no annotations we still compute OKS on it right?

JSabadin: I was thinking about this. We can't calculate OKS for images without GT because it would affect the final result. In other implementations, such patterns are skipped, so we I'll fix this.

klemen1999 (approved):

PR #290, opened 2025-08-28
Fix annotate scenario where paths have different lengths

JSabadin:
## Purpose

In luxonis-ml, metadata strings are converted to lists of integers. PyTorch’s default collate fails when path lengths differ, so this is fixed by a new collate_fn used when annotation is enabled.

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


kozlov721:
Could you please take a look at the failing test? It looks like the `metadata_task_override` config node parameter is not used correctly during `annotate`.

PR #358, opened 2026-03-17
Advanced Finetuning

kozlov721:
## Purpose

Improves the control over the training process: per-node parameter groups with their own optimizer/scheduler settings, and robust node freezing/unfreezing.

### User-facing feature

- Adds new field `finetuning` to the `NodeConfig`
- This field allows users to specify different parameter groups used for optimization
  - Allows to set different optimizer/scheduler parameters or different optimizers/schedulers altogether for different parameter groups of each node
  - For example allows to set lower learning rate for backbone vs head
  - Selectors are case-insensitive, unanchored regexes over parameter names and module class names; entries are first-match-wins; omitted optimizer/scheduler names inherit from the trainer level (see the *Behavior notes* in `configs/README.md`)
- Added `finetuning` argument to `SimplePredefinedModel`
- Updated `ocr_recognition_model.yaml` to use different `weight_decay` for `nn.Linear` modules in the `OCRCTCHead` as per the [reference implenentation](https://github.com/PaddlePaddle/PaddleOCR/blob/main/ppocr/modeling/heads/rec_ctc_head.py)

**Example:**

```yaml
nodes:
  - name: CustomHead

    # Specifies custom value of `weight_decay` for all parameters
    # of `nn.Linear` modules in this head. Additionally sets
    # different optimizer and scheduler for the `aux_head` module. 
    finetuning:
      - parameters: keypoint_heads
        optimizer:
          name: Adam
          params:
            lr: 0.0042
        scheduler:
          name: ReduceLROnPlateau
          params:
            mode: max
      - parameters:
          module_type: Conv2d
        optimizer:
          params:
            weight_decay: 0.0001

# Optimizers and schedulers that are used for the rest of the model
trainer:
  optimizer:
    name: Adam
  scheduler:
    name: StepLR
```

### Architecture (see the walkthrough comment for the full reviewer's guide)

After an in-depth review of earlier revisions surfaced several structural problems (checkpoint-resume corruption, a silently-untrained node under `training_strategy` + `freezing`, gradient accumulation being disabled, dependence on Lightning-internal hook ordering), the implementation was redesigned around four ideas:

- **One optimizer object, always.** All optimizer configurations are resolved into inner optimizers of a single `CompositeOptimizer` (`luxonis_train/optimizers/composite_optimizer.py`), so the model stays in Lightning's **automatic optimization** permanently. Gradient accumulation, trainer-level (global-norm) gradient clipping, mixed precision, and checkpointing all go through Lightning's standard paths. Configs that resolve to a single optimizer bypass the composite entirely and keep byte-identical checkpoints and metric names.
- **A total, static parameter partition.** `luxonis_train/lightning/training_plan.py` resolves node `finetuning` rules, strategy rules, and a default tail into a partition where *every* parameter (frozen ones included) belongs to exactly one group. Unclaimed parameters can no longer silently end up without an optimizer.
- **Declarative freezing.** Freezing/unfreezing is a pure function of the config and the epoch (`luxonis_train/lightning/freezing.py`), applied idempotently — group membership never changes, so checkpoint resume is a plain `state_dict` round trip with no custom callback state. `lr_after_unfreeze` rebases the group's *base* learning rate (group `lr`, `initial_lr`, scheduler `base_lrs`), giving deterministic semantics for every scheduler family.
- **Strategies as rule providers.** `BaseTrainingStrategy` v2 contributes parameter-group rules evaluated between node rules and the default tail; `TripleLRSGDStrategy` is ported with identical warmup numerics. Strategies written against the old `configure_optimizers` API are mounted through `LegacyStrategyAdapter` with a deprecation warning.

**Breaking change:** `OCRCTCHead` no longer accepts the `fc_decay` parameter (the previous implementation set a non-functional `regularizer` attribute — see the removed `TODO` in `luxonis_train/nodes/heads/ocr_ctc_head.py`). Because `NodeConfig` uses `extra=forbid`, configs still passing `fc_decay` in the head's `params` will now fail schema validation. Migrate to the new `finetuning` block with `module_type: Linear` and a `weight_decay` override — see `configs/ocr_recognition_light_model.yaml` for the pattern.

**Breaking change:** `BaseTrainingStrategy` moves to a rule-based API (`rules()` + `get_base_configs()` + optional `update_parameters()` over group handles). Out-of-tree strategies implementing the released `configure_optimizers()` contract keep working through a compatibility adapter (with a deprecation warning) until the next minor release; an authoring/migration guide was added to `luxonis_train/strategies/README.md`.

Additional notes:

- The multi-optimizer *manual-optimization* mode present in earlier revisions of this PR (never released) is gone; `automatic_optimization` is always `True`. Checkpoints produced by those earlier revisions with more than one optimizer will not load into the composite (a clear error is raised instead of a silent partial load). Checkpoints from released versions are unaffected.
- Under `TripleLRSGDStrategy`, parameters that are not literal `weight`/`bias` module attributes (e.g. scalar `nn.Parameter`s) previously received **no optimizer**; they now flow to the default tail. The strategy's dead `defaults["momentum"]` warmup writes (which never reached the parameter groups) were removed — training numerics are unchanged.
- Known documented limitation (unchanged from Lightning's own `BaseFinetuning`): `freezing.unfreeze_after` is not supported with multi-GPU DDP, because DDP registers its gradient-sync hooks at wrap time.

## Deployment Plan

None / not applicable

## Testing & Validation

- `tests/unittests/test_finetuning/` (~110 tests): selectors, inheritance, grouping matrix, freezing, the `CompositeOptimizer`/composite-scheduler contract (Lightning `Optimizable` protocol, `LightningOptimizer` wrapping, writable state view, versioned checkpoint format), strategy rules, and the legacy adapter.
- Real-fit end-to-end tests: multi-optimizer training with gradient clipping, **resume from checkpoints taken before and after the unfreeze epoch**, gradient accumulation with two inner optimizers, `16-mixed` precision smoke test, and the `training_strategy` + `finetuning` + `freezing` regression (the previously silently-untrained node now trains, `10/10` parameters).
- `tests/unittests/test_strategies/`: the golden 15-case TripleLR warmup/schedule numeric table is preserved verbatim.
- Full unit suite, pyright gate, and pre-commit hooks pass.

## Summary by CodeRabbit

* **New Features**
  * Per-node finetuning: select parameter groups and override optimizer/scheduler per node; improved multi-optimizer support with manual-optimization handling and optimizer/scheduler logging.

* **Bug Fixes**
  * Removed implicit T_max auto-population for CosineAnnealingLR.
  * Improved ReduceLROnPlateau stepping and metric monitoring in manual mode.

* **Configuration**
  * Removed legacy fc_decay parameter; finetuning config supports weight-decay overrides.

* **Documentation**
  * Added finetuning docs and YAML examples.

* **Tests**
  * Extensive finetuning unit tests covering selectors, grouping, inheritance, errors, and strategies.


dtronmans (commented):
These are my comments so far regarding potential breaking changes or unexpected behavior, I will just take a quick look at the changes to `ReduceLROnPlateau` to note any changes to the scheduler

dtronmans (approved):
Looked through the previous comments and tried the cases with the new changes and everything looks good, and the correct optimizers are logged for the correct parameter groups

luxonis_train/lightning/utils.py
```diff
@@ -265,6 +337,139 @@ def _override_metadata_labels(
...
+        used_params = set(used_params or set())
+        if include_default and not any(
+            node.finetuning for node in self.values()
+        ):
+            params = []
+            for node in self.values():
+                for module in node.module.modules():
+                    if list(module.parameters()) and not list(
+                        module.children()
+                    ):
```
klemen1999: This only collects parameters from leaf modules, so direct parameters on a module that also has children are skipped. 
That misses parameters owned directly by a container module that also contains submodules. For example:
```
class BottleRep(nn.Module):
    def __init__(...):
        self.conv_1 = ...
        self.conv_2 = ...
        self.alpha = nn.Parameter(torch.ones(1))
```
`BottleRep` has children (conv_1, conv_2), so it is skipped. Its children are visited and their convolution weights are added, but alpha belongs to `BottleRep` itself and is never seen.

Result: `alpha` participates in forward/backward, gets a gradient, but is in no optimizer parameter group, so `optimizer.step()` never updates it. This also affects custom nodes with trainable positional embeddings, scaling factors, learned tokens, etc. 
**The same leaf-only logic exists in both the no-finetuning path and the finetuning selector path. (line 415 in this file)**

Proposed fix by AI:
```
for module_name, module in node.module.named_modules():
    for param_name, p in module.named_parameters(recurse=False):
        if id(p) in used_params:
            continue

        name = f"{module_name}.{param_name}" if module_name else param_name
        ...
```
Use this in both optimizer-collection paths. recurse=False is important: it includes direct parameters of both parents and leaves, without repeatedly collecting child parameters. Keep the existing `used_params` identity check as a safeguard.

luxonis_train/lightning/luxonis_lightning.py
```diff
@@ -577,21 +656,79 @@ def on_save_checkpoint(self, checkpoint: dict[str, Any]) -> None:
...
+                for p in group["params"]
+            }
+            strategy_optimizers, strategy_schedulers = (
+                self.training_strategy.configure_optimizers(used_params)
+            )
+            optimizers = [*optimizers, *strategy_optimizers]
+            schedulers = [*schedulers, *strategy_schedulers]
+        else:
+            optimizers, schedulers = self.nodes.build_optimizers()
+        if len(optimizers) > 1:
```
klemen1999: The PR supports configurations that create multiple optimizers, for example:
```
finetuning:
  - parameters: backbone
    optimizer:
      name: Adam
      params: {lr: 0.0001}
  - parameters: head
    optimizer:
      name: SGD
      params: {lr: 0.01}
```
It switches Lightning into manual optimization and steps every optimizer once per batch.
Separately, the existing `TrainingManager` handles `freezing` by adding newly unfrozen node parameters to an optimizer at the target epoch. Lightning invokes this callback once for each optimizer. The callback’s logic has no guard to select the owning optimizer, so:
- First optimizer: the formerly frozen node is unfrozen and added.
- Second optimizer: the same node is added again, because the callback is invoked again.
- Every subsequent optimizer: same thing.

After that, the manual training loop calls `step()` on each optimizer. Those formerly frozen parameters are updated multiple times per batch—once with every optimizer’s algorithm and learning rate. They also lose the intended finetuning rule; e.g. a frozen backbone configured for Adam can be added to the head’s SGD optimizer too.
This is specifically introduced by combining the new multi-optimizer support with existing freezing. The implementation needs either to reject that combination or store a parameter-to-optimizer ownership map and add unfrozen parameters only to their assigned optimizer/group.

Proposed fix by AI:
I’d avoid `BaseFinetuning.unfreeze_and_add_param_group` for the new parameter-group system.

Recommended approach:

1. Build the parameter-to-optimizer/group assignment independently of `requires_grad`. A frozen parameter should still be assigned to exactly one planned optimizer group.

2. Include frozen parameters in their intended groups when optimizers are created. PyTorch permits parameters with `requires_grad=False`; they simply have no gradient and are not updated.

3. Change `TrainingManager`’s unfreeze path to only set those parameters back to `requires_grad=True`. It must not call `add_param_group`.

4. Run unfreezing once per epoch, not once per optimizer. Override a callback epoch hook or track completed `(node, epoch)` pairs, rather than relying on `BaseFinetuning.finetune_function`, which Lightning calls for every optimizer.

5. If `lr_after_unfreeze` is retained, make frozen-node groups distinct and set that group’s LR at unfreeze time. This prevents changing the LR of unrelated parameters that happened to share an optimizer.

This gives each parameter one stable owner throughout training, including while frozen. It also fixes a related edge case: an entirely frozen finetuning rule currently has no `requires_grad=True` parameters, so it cannot form its intended optimizer group at setup time.

At minimum, if this refactor is too large for the PR, reject configurations combining `freezing.active` with multiple optimizers. That is safer than silently adding the same parameters to every optimizer.

klemen1999 (commented):
2 bugs found by AI tools

kozlov721:
## Reviewer's guide to the redesign

This branch grew a lot between the initial implementation and now, so here is a map of *why* it changed and *how to read it*. TL;DR: an in-depth review (plus an adversarial verification pass with runnable repros) found that the defects in the original approach were structural, not local — so instead of patching symptoms, the optimizer/freezing machinery was rebuilt around a small set of invariants, and the diff is net-negative in the core (`-~600` lines of workarounds).

### What was wrong (all reproduced, not theoretical)

| Finding | Symptom |
| --- | --- |
| Resume after unfreeze | `BaseFinetuning`'s checkpointed metadata went stale when unfreezing mutated param groups in place → resume either crashed in `Optimizer.load_state_dict` ("parameter group that doesn't match the size") or restored **0-parameter optimizers** and trained nothing, silently. Resuming *past* the unfreeze epoch re-froze the node forever (`e == epoch` gate + unconditional re-freeze at `setup`). |
| `training_strategy` + `finetuning` + `freezing` | The frozen node's parameters ended up in **no optimizer at all** after unfreezing (measured `0/10`), silently — the strategy filtered `requires_grad` at build time and the id-exclusion contract couldn't adopt parameters later. |
| Manual-optimization mode (>1 optimizer) | Lightning's guardrails were skipped only because `_verify_loop_configurations` happens to run before `configure_optimizers`; a second `fit()` on the same model crashed with `MisconfigurationException`. Gradient accumulation was disabled; per-optimizer clipping made the model-wide grad norm `√N ×` the single-optimizer norm; scheduler stepping had to mirror Lightning's internals by hand. |
| `lr_after_unfreeze` | Only `group["lr"]` was written → schedule-dependent behavior (`LambdaLR` clobbered it on the next step, `StepLR` kept it, `CosineAnnealingLR` partially). |
| Grouping identity | Scheduler defaults (`CosineAnnealingLR.T_max`) were injected *after* grouping keys were computed → semantically identical schedulers failed to merge → spurious extra optimizer → spurious manual mode. |

### The design inversions

1. **One optimizer object, always.** `CompositeOptimizer` exposes the inner optimizers' `param_groups` (the *same* dict objects, concatenated) behind a single `torch.optim.Optimizer`, so Lightning stays in **automatic optimization** permanently. Everything class-B above disappears: accumulation and trainer-level clipping work again (clipping is now one global norm over all groups, like any single-optimizer run), AMP goes through the one GradScaler slot, and there is no hook-order coupling left. Single-optimizer configs bypass the composite — plain configs keep byte-identical checkpoints and `lr-Adam` metric keys.
2. **Total, static partition.** `resolve_training_plan` claims *every* parameter (frozen included) into exactly one group: node rules (YAML order) → strategy rules → a default tail. This kills the "no optimizer for the unfrozen node" bug **by construction** — there is no code path that can leave a parameter unowned.
3. **Freezing is `requires_grad`, nothing else.** Group membership never changes; torch optimizers natively skip grad-less parameters. Unfreeze = flip the flag (idempotently re-derived from the schedule at every epoch start), so resume is a plain `state_dict` round trip with **zero custom checkpoint state**. `lr_after_unfreeze` rebases group `lr` + `initial_lr` + the member scheduler's `base_lrs[i]` — deterministic for every scheduler family (documented table in `configs/README.md`).
4. **Strategies contribute rules** instead of building optimizers behind an id-exclusion set. `TripleLRSGD` becomes three structural rules (BN-weights / weights / biases) with tag-based warmup writes; the golden numeric warmup table is preserved verbatim. Old-API strategies mount through `LegacyStrategyAdapter` (deprecation warning; overlap with finetuning rules is a hard error).

### Reading order (each commit leaves the suite green)

| Commit | What to review |
| --- | --- |
| `03e9e3c` | Hotfixes for the reviewable-in-isolation findings (kept as the baseline; parts are superseded later). |
| `c31ec38` | The new modules, unwired: `optimizers/composite_optimizer.py`, `schedulers/composite_scheduler.py`, `lightning/training_plan.py`, `lightning/freezing.py` + their contract tests. **This is where the invariants live** — review hardest here. |
| `567eb8a` | Non-strategy path wired through the plan; `TrainingManager` rewritten as a plain callback (drops `BaseFinetuning` and all checkpointed state). |
| `1829b06` | Strategy API v2 + TripleLR port + legacy adapter; deletion of the entire manual-optimization machinery (net `+722/−1491`). |
| `5cc85a8` | End-to-end coverage: strategy+freezing regression, resume matrix, accumulation, `16-mixed` smoke, plateau/epoch scheduler coexistence, legacy adapter. |
| `d30b75d` | Docs (`configs/README.md` behavior notes, `strategies/README.md` authoring + migration guide). |

### What is deliberately unchanged

- The **YAML surface** and its documented semantics: selector syntax, first-match-wins, unanchored case-insensitive matching, inheritance/merging of optimizer/scheduler overrides, the "entries sharing optimizer+scheduler collapse into one optimizer with one param group per entry" contract (all pinned by the pre-existing structural tests, which still pass against the inner optimizers via a small snapshot-helper unwrap).
- Scheduler timing: Lightning itself steps epoch-interval non-plateau schedulers before validation and plateau schedulers after it — since Lightning drives everything now, parity holds definitionally.
- `LearningRateMonitor` metric keys for existing configs (param groups intentionally carry no `"name"` keys — injecting them would have renamed every logged `lr-*` series).

### Verification evidence

- Resume past the unfreeze epoch, real 2-run fit: previously `ValueError` on restore (or a permanently frozen node); now `requires_grad=10/10, in_optimizer=10/10, moved_vs_ckpt=8` for the previously frozen head.
- `TripleLRSGDStrategy` + finetuning rule + frozen node: previously `in_optimizer=0/10` after unfreezing (silent); now `10/10` and the node measurably trains.
- `accumulate_grad_batches=2` with two inner optimizers: `global_step == n_batches // 2` (previously the setting was ignored with a warning).
- `16-mixed`: one GradScaler `unscale_`/`step` per training step against the composite — indistinguishable from a plain optimizer.
- Full unit suite (373 passed + subtests), pyright gate (baseline only), pre-commit clean; the composite's torch/Lightning contract (protocol attributes, `LightningOptimizer` dynamic subclassing, `_optimizer_to_device` state writes, versioned `state_dict`) is pinned by dedicated tests.

### Known limitation (documented, pre-existing)

`freezing.unfreeze_after` + multi-GPU DDP: DDP registers gradient-sync hooks at wrap time, so parameters unfrozen mid-training aren't synchronized. Same limitation as Lightning's own `BaseFinetuning`; called out in the docs rather than papered over.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

luxonis_train/lightning/training_plan.py
```diff
@@ -0,0 +1,753 @@
...
+    """
+    inner_optimizers: list[Optimizer] = []
+    # one entry per inner: (member scheduler or None, plateau monitor
+    # or None)
+    entries: list[
+        tuple[LRScheduler | ReduceLROnPlateau | None, str | None]
+    ] = []
+    bypass_configs: list[Any] | None = None
+    for inner in plan.inners:
+        torch_groups = [
```
dtronmans, line 658: One last thing I noticed actually, but I sometimes use the `LearningRateMonitor` callback to make sure that `resume_training` loads the optimizer/scheduler/LR state correctly, and also to see if the LR is too low or not at a certain epoch. With the advanced finetuning setup, I see multiple MLFlow logging keys for the learning rate but they are not named in a systematic way so it's difficult to see what LR belongs to what group:

Image

If I change the dict to 

```
 {
          "params": list(group.parameters),
          "name": group.name,
          **group.options,
 }
```

and change the below `optimizer_keys = set(optimizer.defaults) | {"params"}` to `optimizer_keys = set(optimizer.defaults) | {"params", "name"}`

Then I get a named LR per group:

Image

dtronmans (approved):
LGTM and works well, just one thing I just noticed regarding naming param groups for MLFlow logging keys

luxonis_train/lightning/training_plan.py
kozlov721: Good catch, thanks — fixed in 3cdb0f1 on this branch. The plan already names
  every group; the name just never reached the torch parameter group.

  Two adjustments:
  
  1. `name` is set **after** `**group.options`, so the plan's name always wins
     over a stray `name` in a rule's optimizer params.
  2. Names are made unique across the whole plan, not per inner. Lightning raises
     `MisconfigurationException` on two groups with the same `name` in one
     optimizer, and `CompositeOptimizer` presents every inner's groups as one flat
     list — so without a guard this could turn a naming nicety into a crash.
     Collisions get a `-2` suffix.
  
  Same run, before and after (4 groups across 2 inner optimizers):


lr-keys-before-after

  Naming is `<node>/<index>`, `strategy/<tag>`, or `default`, with the node
  appended when a group is node-scoped (`default/Head`); documented in
  `configs/README.md`.

luxonis_train/lightning/utils.py
```diff
@@ -371,174 +456,65 @@ def build_training_strategy(
...
-
-    optimizer = from_registry(
-        OPTIMIZERS,
-        cfg_optimizer.name,
-        **cfg_optimizer.params,
-        params=[p for p in parameters if p.requires_grad],
+    if training_strategy is None:
+        return None
+    logger.info(f"Using training strategy '{training_strategy.name}'")
+    # `optimizer` and `scheduler` have defaults, so they are never
```
klemen1999: Nitpick: Feels like a comment that is related to somethin that was here before and describes why is there now. Can we remove/shorten the comment?

luxonis_train/lightning/training_plan.py
dtronmans: Looks great, and thanks for the diagram! I don't have any other comment regarding this PR so everything LGTM

luxonis_train/lightning/utils.py
```diff
@@ -371,174 +456,65 @@ def build_training_strategy(
...
-            dirpath=save_dir / "min_val_loss",
-            filename=f"{model_name}_loss={{val/loss:.4f}}_{{epoch:02d}}",
-            monitor="val/loss",
-            auto_insert_metric_name=False,
-            save_top_k=cfg.trainer.save_top_k,
-            mode="min",
-        )
-    )
-    return callbacks
+def _raise_not_implemented(self: Any) -> Any:
```
klemen1999: This function has one usage. Do we really need it as a function?

luxonis_train/strategies/triple_lr_sgd.py
```diff
@@ -1,119 +1,51 @@
...
-class TripleLRScheduler:
-    optimizer: Optimizer
-    warmup_epochs: int
-    warmup_bias_lr: float
-    warmup_momentum: float
-    lre: float
-    cosine_annealing: bool
-    epochs: int
-    max_stepnum: int
+def _is_batch_norm_weight(
```
klemen1999: Nitpick: let's put this below class definition

klemen1999 (approved):
Generally LGTM other than one nitpick
I added frozen params also in the reported statistics (in https://github.com/luxonis/luxonis-train/pull/358/commits/25ed97521f8ef65828d4150846fbaa56e6a34814). Example:
image

PR #365, opened 2026-04-02
Map predefined per_class_metrics to metric-specific kwargs

dtronmans:
## Purpose
There are three metrics so far that consume `per_class_metrics`:
- `MeanAveragePrecision`: the Torchmetrics class has the `class_metrics` option for `MeanAveragePrecisionSegmentation` and `MeanAveragePrecisionBBox` (`MeanAveragePrecisionKeypoints` does not inherit from `Torchmetrics` or expose anything class-specific)
- `MIoU` exposes the equivalent `per_class`

The solution to some metrics having per-class metrics and some here is to add a `get_predefined_model_params_aliases` so metrics can declare how `per_class_metrics` should be forwarded. If no metric supports a per-class flag and the user is trying to override `per_class_metrics` then a warning is logged instead of crashing.

Additional metrics-related changes:

- fixes to the README (per_class_metrics was not by default True for segmentation and detection) and addition of `MIoU` (was missing from the README).
- When `per_class` is set to True (passed from `per_class_metrics`) MIoU returns the same main IoU and per-class submetrics

Example config with object detection:

```
model:
  name: detection_light
  predefined_model:
    name: DetectionModel
    variant: light
    params:
      per_class_metrics: true
      task_name: vehicles
      metrics_params:
      loss_params:
        iou_type: "siou"
        iou_loss_weight: 20
        class_loss_weight: 8
```

Result:

per_class_detection

Example config with classification (unsupported argument):

```
model:
  name: classification_light
  predefined_model:
    name: ClassificationModel
    variant: light
    params:
      per_class_metrics: true
      task_name: vehicles
```

Result:

classification

Example config with segmentation:

```
model:
  name: segmentation_light
  predefined_model:
    name: SegmentationModel
    variant: light
    params:
      task_name: vehicles
      per_class_metrics: true
      metrics_params:
        num_classes: 3
```

Result:

segmentation_metrics

### DIFF
--- a/luxonis_train/config/predefined_models/base_predefined_model.py
+++ b/luxonis_train/config/predefined_models/base_predefined_model.py
@@ -1,6 +1,7 @@
 from abc import abstractmethod
 from typing import Literal
 
+from loguru import logger
 from luxonis_ml.typing import Kwargs, Params, check_type
 from typeguard import typechecked
 from typing_extensions import override
@@ -12,7 +13,7 @@ from luxonis_train.config import (
     NodeConfig,
 )
 from luxonis_train.config.config import FreezingConfig
-from luxonis_train.registry import MODELS
+from luxonis_train.registry import METRICS, MODELS, NODES
 from luxonis_train.variants import VariantBase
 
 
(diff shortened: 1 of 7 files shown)


luxonis_train/config/predefined_models/base_predefined_model.py
```diff
@@ -142,9 +141,52 @@ def __init__(
         )
         self._confusion_matrix_params = confusion_matrix_params or {}
 
+    def _generate_metrics(self) -> list[MetricModuleConfig]:
```
klemen1999: Nitpick: Let's put private methods below public ones

klemen1999 (approved):

PR #366, opened 2026-04-08
AIMET Integration

kozlov721:
## Purpose

Adds option to quantize trained model using various PTQ and QAT techniques.

## Specification

- Added `LuxonisModel.quantize` method
  - Runs selected algorithms on the model weights and exports it to ONNX
  - Support for various PQT techniques:
    - Adaround
    - Batch norm folding
    - Cross-layer equalization
    - Batch norm re-estimation
    - Sequential MSE
- Added `AIMETCallback`
  - Runs quantization at the end of the training 
- Added a new section `exporter.aimet` to the `Config`
- Other small changes necessary to accomodate the AIMET API or to simplify the integration
  - Added more standard `forward` method to `LuxonixLightningModule`
    - The original renamed to `full_forward`
  - Support for loaders outputting bare tensors
    - As opposed to the expected dictionary
  - Removed metrics, losses and visuazilers from saved node modules
  - Added `print_table` as a required abstract method to `BaseLuxonisProgressBar`
  - Implemented custom `__getstate__` and `__setstate__` in `LuxonisLightningModule`

**`config.yaml`**
```yaml
# Default values
exporter:
  aimet:
    active: false

    default_output_bw: 8
    default_param_bw: 8
    default_data_type: "int"
    quant_scheme: "min_max"
    config: ~

    fold_batch_norms: false
    cross_layer_equalization: false
    batch_norm_reestimation: false
    sequential_mse: false
    adaround:
      active: false
      default_num_iterations: ~
      default_reg_param: 0.01
      default_beta_range: [20, 2]
      default_warm_start: 0.2

    epochs: 20
    optimizer:
      name: "SGD"
      params:
        lr: 0.00001
    scheduler:
      name: "StepLR"
      params:
        step_size: 5
        gamma: 0.1

```

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

- Testing `AIMETCallback` in `test_callbacks`
- Testing `LuxonisModel.quantize` for all predefined models with the full set of PTQ techniques enabled

## Summary by CodeRabbit

* **New Features**
  * AIMET quantization: new quantize CLI command and model.quantize API; optional AIMET install instructions included.

* **Bug Fixes & Improvements**
  * Loaders now accept raw tensors or image-keyed dicts.
  * Improved export/pickling, progress-bar table rendering, visualization stability, and training callback behavior.
  * Model initialization accepts weights from the CLI.

* **Documentation**
  * Installation and CLI docs updated to document AIMET/quantize.


kozlov721:
Draft for now because the keypoint and anomaly detection models don't work yet.

Some open questions:
- Do we want to auto-enable the AIMETCallback for all predefined models?
- Which of the PTQ techniques do we want to have enabled by default?
- Should the config `aimet` section be under `exporter` or on its own?

klemen1999:
> Some open questions:
> 
> * Do we want to auto-enable the AIMETCallback for all predefined models?

For now no. Maybe later on when we experiment a bit more with this method.
> * Which of the PTQ techniques do we want to have enabled by default?

What does AIMET use for default? I have seen Adaround before (Hailo) so my first bet would be to use that as default but I haven't really looked at other ones yet.
> * Should the config `aimet` section be under `exporter` or on its own

Let's keep it under exporter, I think this is fine

kozlov721:
> What does AIMET use for default?

These are all individual AIMET APIs, none of them is really used "by-default" anywhere. But I saw the batch norm folding being used a lot. Adaround seems like a good method and is used by the hailo exporter too but can be pretty slow (although probably alright compared to the lenght of the entire training run).

Adaround with the example toy coco dataset on the detection light model took about 40 minutes I think when I was testing it. So with a large dataset and on a larger model it can easily take multiple hours.

kozlov721:
> So with a large dataset and on a larger model it can easily take multiple hours.

This depends on the `adaround.default_num_iterations` parameter, but in general Adaround needs a lot of iterations to converge. By default AIMET uses 15,000 iterations per layer.

klemen1999:
> > What does AIMET use for default?
> 
> These are all individual AIMET APIs, none of them is really used "by-default" anywhere. But I saw the batch norm folding being used a lot. Adaround seems like a good method and is used by the hailo exporter too but can be pretty slow (although probably alright compared to the lenght of the entire training run).
> 
> Adaround with the example toy coco dataset on the detection light model took about 40 minutes I think when I was testing it. So with a large dataset and on a larger model it can easily take multiple hours.

What if we just make this a required parameter is PTQ is active? So that we don't need to select some default as well? Otherwise I'm fine going with batch norm folding as well for default. Adaround is probably a bit too overkill to have it for default case yeah.

kozlov721:
> What if we just make this a required parameter is PTQ is active?

You mean the opt-in advanced techniques? Yeah I think this would make sense. What you want to use usually also depends on the specific model so as long as the entire AIMET is opt-in I think forcing the user to explicitly specify what to use is a good idea. 

I suppose eventually we will learn what works best with our predefined models and we can then enable it by default for them with the known well working parameters.

configs/README.md
```diff
@@ -510,6 +510,7 @@ Here you can define configuration for exporting.
 | `onnx`                   | `dict`                | `{}`              | Options specific for ONNX export. See [ONNX](#onnx) section for details                        |
 | `hubai`                  | `dict`                | `{}`              | Options for HubAI SDK conversion. See [HubAI](#hubai) section for details                      |
 | `blobconverter`          | `dict`                | `{}`              | Options for converting to BLOB format (deprecated). See [Blob](#blob-deprecated) section       |
+| `aimet`                  |                       | `dict`            | `{}`                                                                                           |
```
dtronmans: I think the columns are moved one-to-the-right accidentally, it should be 

| `aimet`                  |`dict`                       |`{}`             | description                                                                                         |

configs/README.md
```diff
@@ -571,6 +572,41 @@ exporter:
...
+| `default_data_type`        | `Literal["int", "float"]`                         | `int`         | Default data type for quantized values                                                                                                                                                                                                                                  |
+| `quant_scheme`             | `Literal["min_max", "post_training_tf_enhanced"]` | `min_max`     | Quantization scheme to use                                                                                                                                                                                                                                              |
+| `config`                   | `dict \| str`                                     | `{}`          | Additional configuration for AIMET. Can be a dictionary or a path to a JSON config file. Refer to the [AIMET documentation](https://quic.github.io/aimet-pages/releases/latest/docs/QuantizationSim.html#quantization-sim-config) for details on the available options. |
```
dtronmans: I think the link might be broken, when I open it I get 404

luxonis_train/core/core.py
```diff
@@ -1242,6 +1282,245 @@ def convert(
...
+            )
+        if scheduler is None:
+            scheduler = from_registry(
+                SCHEDULERS,
+                cfg.scheduler.name,
+                optimizer=optimizer,
+                **cfg.scheduler.params,
+            )
+
+        model = quantization_aware_training(
```
dtronmans, line 1556: Is there a way to disable QAT and only run PTQ in the config? Right now it seems as though QAT is unconditional

luxonis_train/config/config.py
```diff
@@ -725,6 +727,87 @@ def _validate_quantization_mode(value: str) -> str:
...
+
+    fold_batch_norms: bool = False
+    cross_layer_equalization: bool = False
+    batch_norm_reestimation: bool = False
+    sequential_mse: bool = False
+    adaround: AdaroundConfig = Field(default_factory=AdaroundConfig)
+
+    epochs: PositiveInt = 20
+    optimizer: ConfigItem = Field(
+        default_factory=lambda: ConfigItem(name="SGD", params={"lr": 1e-5})
```
dtronmans, line 757: What if the `trainer.optimizer.SGD.lr` is set in the general (trainer) config, but the `exporter.aimet.optimizer` is not set? Then this default 1e-5 will be written instead of inheriting form the (standard) `trainer.optimizer.SGD.lr` but the README says that "If not set, the `trainer` optimizer is used"

Same for the scheduler fields `step_size` and `gamma` below

configs/README.md
```diff
@@ -571,6 +572,41 @@ exporter:
...
+| `default_param_bw`         | `int`                                             | `8`           | Default bitwidth for quantized parameters                                                                                                                                                                                                                               |
+| `default_data_type`        | `Literal["int", "float"]`                         | `int`         | Default data type for quantized values                                                                                                                                                                                                                                  |
+| `quant_scheme`             | `Literal["min_max", "post_training_tf_enhanced"]` | `min_max`     | Quantization scheme to use                                                                                                                                                                                                                                              |
```
dtronmans: The quant_scheme values mentioned in the config are `quant_scheme: Literal["min_max", "tf", "tf_enhanced"] = "min_max"`

luxonis_train/core/core.py
```diff
@@ -1242,6 +1282,245 @@ def convert(
 
         return archive_path, conversion_artifacts
 
+    def quantize(
+        self,
+        weights: PathType | None = None,
+        epochs: int | None = None,
+        quant_scheme: Literal["min_max", "tf", "tf_enhanced"] | None = None,
+        default_output_bw: int | None = None,
```
dtronmans, line 1348: Isn't the default for these values, from the config:

`default_output_bw`: 8
`default_param_bw`: 8

luxonis_train/config/config.py
```diff
@@ -725,6 +727,87 @@ def _validate_quantization_mode(value: str) -> str:
...
+            name="StepLR", params={"step_size": 5, "gamma": 0.1}
+        )
+    )
+
+    @model_validator(mode="before")
+    @classmethod
+    def validate_active(cls, data: Params) -> Params:
+        if not data.get("active", False):
+            return data
+        for required_field in [
```
dtronmans: I think that for some parameters there is a mismatch between the validator, the config and the signature in the `quantize` method. Because `fold_batch_norms` in the config is by default False, but here it is checking whether or not it is set (even though it doesn't necessarily need to be set since there is a default value). And in the `quantize()` method the `fold_batch_norms` has default `None` in the signature

requirements.txt
```diff
@@ -19,7 +19,7 @@ semver~=3.0
 tabulate~=0.9
 tensorboard~=2.20
 termcolor~=3.2
-torch<2.11 # temporary pin: GitHub runner image does not yet support the newer NVIDIA driver requirement used by torch 2.11 wheels
```
dtronmans, line 22: Could you please keep the comment? I think the torch requirement can soon be moved upwards if we confirm that the runner image supports CUDA 12.8/13

dtronmans (commented):
Just a few comments from reading it (did not test locally yet), could you also please include in the PR comment a few notes/justifications on the changes that are not necessarily related to the PR? This would be related to the changes to the losses, backbones, visualizers etc... Just to know what is in the scope of AIMET and what is not

luxonis_train/config/config.py
kozlov721: >"If not set, the trainer optimizer is used"

This is an old incorrect note, thanks. 

The QAT has a separate optimizer (scheduler) because it usually requires specific hyperparams that are probably different from what you used for the main training.

configs/README.md
kozlov721: `"min_max"` and `"tf"` is the same, `"tf"` is a deprecated name so I'm not mentioning it in the README but I made the config still accept it.

luxonis_train/core/core.py
kozlov721: It needs to be set to `None` by default here because we need to distuinguish between when the user sets the option in a config or in the function arguments (which would take precedence).

kozlov721: If not set the values stay `None` and the defaults will be propagated from the config (where it's either set to the default `8` or to whatever the user chose).

luxonis_train/core/core.py
kozlov721: You can disable it by setting the number of epochs to 0.

luxonis_train/config/config.py
kozlov721: This is intentional:

> > What if we just make this a required parameter is PTQ is active?
> 
> You mean the opt-in advanced techniques? Yeah I think this would make sense. What you want to use usually also depends on the specific model so as long as the entire AIMET is opt-in I think forcing the user to explicitly specify what to use is a good idea.

We decided that for the near future if you enable `AIMET` you have to explicitly enable or disable each technique you want to use.

luxonis_train/attached_modules/losses/efficient_keypoint_bbox_loss.py
```diff
@@ -74,9 +72,7 @@ def __init__(
             **kwargs,
         )
 
-        self.b_cross_entropy = BCEWithLogitsLoss(
-            pos_weight=torch.tensor([viz_pw])
-        )
+        self.pos_weight = torch.tensor([viz_pw])
```
kozlov721, line 75: Keeping a reference to another `BaseLoss` that is not attached to a specific node is problematic when copying the model. In this case it's easier to use `F.binary_cross_entropy_with_logits` directly instead of our `BSEWithLogitsLoss`.

luxonis_train/attached_modules/losses/efficient_keypoint_bbox_loss.py
```diff
@@ -264,12 +270,3 @@ def dist2kpts_noscale(self, anchor_points: Tensor, kpts: Tensor) -> Tensor:
         adj_kpts[..., 0] += x_adj
         adj_kpts[..., 1] += y_adj
         return adj_kpts
-
-    def _init_parameters(self, features: list[Tensor]) -> None:
```
kozlov721, line 270: Parameters created this way were causing weird errors during quantization because the tensors were created during `inference_mode`.

luxonis_train/attached_modules/losses/reconstruction_segmentation_loss.py
```diff
@@ -92,7 +92,7 @@ def forward(self, img1: Tensor, img2: Tensor) -> Tensor:
 
             (_, channel, _, _) = img1.size()
             if channel == self.channel and self.window.dtype == img1.dtype:
-                window = self.window.to(device)
+                window = self.window.to(device).clone()
```
kozlov721, line 99: Cloning is another way how to fix the "tensors created in inference mode" issue.

luxonis_train/attached_modules/metrics/base_metric.py
```diff
@@ -164,6 +164,12 @@ def compute(
         """
         return super().compute()
 
+    def __eq__(self, other: object) -> bool:
```
kozlov721, line 168: This fixes an issue with `fold_all_batch_norms`. 

`torchmetrics.Metric` supports chaining individual metrics into larger pipelines using overloaded math operators. 

For example you can do:

```python
from torchmetrics import Precision, Recall

precision = Precision(task="binary")
recall = Recall(task="binary")

# Operator overloading on the classes
f1_score = 2 * (precision * recall) / (precision + recall)
```

The `f1_score` works the same as the official `torchmetrics.F1Score`, but it was created by pipelining smaller metrics. 

```python
from torchmetrics import F1Score

official_f1_score = F1Score(task="binary")

f1_score.update(
    tensor([0, 1, 1, 0, 0, 1]), tensor([0, 1, 0, 1, 1, 1])
)
print(f"F1 Score: {f1_score.compute()}")
# F1 Score: 0.5714

official_f1_score.update(
    tensor([0, 1, 1, 0, 0, 1]), tensor([0, 1, 0, 1, 1, 1])
)
print(f"Official F1 Score: {official_f1_score.compute()}")
# Official F1 Score: 0.5714

print(f1_score)
# CompositionalMetric(
#   true_divide(
#     CompositionalMetric(
#       mul(
#         2,
#         CompositionalMetric(
#           mul(
#             BinaryPrecision(),
#             BinaryRecall()
#           )
#         )
#       )
#     ),
#     CompositionalMetric(
#       add(
#         BinaryPrecision(),
#         BinaryRecall()
#       )
#     )
#   )
# )
```

This is cool but it works for `==` and `!=` as well and for combinations of `torchmetrics.Metric` and other types:

```python
foo = precision != 2
print(foo)
# CompositionalMetric(
#   ne(
#     BinaryPrecision(),
#     2
#   )
# )
```

This has one major disadvantage:
```python
foo = precision != 2
print(bool(foo))
# True

if precision == None:
    print("This should not happen")
# This should not happen
```
This breaks AIMET which expects comparisons to work, so in order to fix it we have to re-implement `__eq__` (and `__hash__`) ourselves.

luxonis_train/attached_modules/visualizers/base_visualizer.py
```diff
@@ -25,6 +26,13 @@ def __init__(self, *args, scale: float = 1.0, **kwargs) -> None:
         super().__init__(*args, **kwargs)
         self.scale = scale
 
+    @override
```
kozlov721, line 30: `ColorMap` internally uses a generator so it cannot be pickled. Before pickling we remove it from the instance state.

luxonis_train/lightning/luxonis_lightning.py
```diff
@@ -191,8 +194,43 @@ def core(self) -> "luxonis_train.core.LuxonisModel":
 
     @override
     def forward(
+        self, inputs: dict[str, Tensor] | Tensor
```
kozlov721, line 239: AIMET expects a simple `forward` that expects a regular `nn.Tensor`.

luxonis_train/lightning/luxonis_lightning.py
```diff
@@ -997,6 +994,20 @@ def get_mlflow_logging_keys(self) -> dict[str, list[str]]:
             "artifacts": sorted(artifact_keys),
         }
 
+    @override
```
kozlov721, line 1053: `trainer` and `core` cannot be pickled which causes the quantization to fail. We could just delete them before pickling but we do need to actually remember their states. This hack makes pickling and unpickling work as long as there is only one instance of `LuxonisLightning` which is good enough for the quantization to succeed.

luxonis_train/lightning/utils.py
```diff
@@ -82,9 +84,9 @@ def __init__(
         super().__init__()
         self.name = name
         self.module = module
-        self.losses = _to_module_dict(losses)
```
kozlov721, line 85: Having the attached modules registered as modules of the overall `LuxonisLightning` module caused issues during quantization. After this change the attached modules are no longer considered a part of the core model itself.

luxonis_train/loaders/base_loader.py
```diff
@@ -224,19 +224,8 @@ def augment_test_image(self, img: dict[str, Tensor]) -> Tensor:
             "`augment_test_image` method to expose this functionality."
         )
 
-    def __getitem__(self, idx: int) -> LuxonisLoaderTorchOutput:
```
kozlov721, line 229: `AIMET` expects dataloaders to have a regular `__getitem__` that returns a tuple with a simple `torch.Tensor` and labels.

kozlov721 (commented):
> could you also please include in the PR comment a few notes/justifications on the changes that are not necessarily related to the PR?

Mostly all the changes are somehow related to making AIMET work. I added more comments to some individual changes in code. If you have more questions about some other specific changes let me know.

dtronmans (approved):

PR #369, opened 2026-04-09
`DatasetMetadata` loading from weights

kozlov721:
## Purpose

When weights are provided to `LuxonisModel`, the `DatasetMetadata` are loaded from it if the weights contain it.

## Specification

- Moved config restoration from checkpoint from `__main__` to `LuxonisModel`

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/luxonis_train/config/config.py
+++ b/luxonis_train/config/config.py
@@ -43,6 +43,7 @@ from typing_extensions import Self, override
 
 import luxonis_train as lxt
 from luxonis_train.registry import MODELS, NODES, from_registry
+from luxonis_train.upgrade import upgrade_config
 
 
 class ImageSize(NamedTuple):
@@ -865,6 +866,9 @@ class Config(LuxonisConfig):
         cfg: PathType | Params | None = None,
         overrides: Params | list[str] | tuple[str, ...] | None = None,
     ) -> "Config":
+        if cfg is not None:
+            cfg = upgrade_config(cfg)
+
         instance = super().get_config(cfg, overrides)
         if not isinstance(cfg, str):
             return instance.smart_auto_populate()
--- a/luxonis_train/config/config.py
+++ b/luxonis_train/config/config.py
@@ -867,15 +867,14 @@ class Config(LuxonisConfig):
             cfg = upgrade_config(cfg)
 
         instance = super().get_config(cfg, overrides)
-        if not isinstance(cfg, str):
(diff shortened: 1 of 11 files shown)


luxonis_train/core/core.py
```diff
@@ -92,16 +93,38 @@ def __init__(
...
+            in the config file, the weights provided here will take
+            precedence.
+
         @type debug_mode: bool
         @param debug_mode: If set to True, enables debug mode which ignores some
             normaly unrecovarable exceptions and allows to test the model
             without it being fully functional.
         """
+        if weights is not None:
+            ckpt = torch.load(weights, map_location="cpu")  # nosemgre
```
klemen1999: Previoulsy we supported `weights` also being remote path. Now this fails because we immediately try to torch.load() thema

luxonis_train/core/core.py
```diff
@@ -319,18 +342,7 @@ def train(
                 self.cfg.trainer.matmul_precision
             )
 
-        if weights is not None:
-            weights = LuxonisFileSystem.download(
-                str(weights), self.run_save_dir
-            )
-            if self.cfg.model.weights is not None:
-                logger.warning(
```
klemen1999, line 327: I'd still keep this warning

luxonis_train/config/config.py
```diff
@@ -865,6 +866,9 @@ def get_config(
         cfg: PathType | Params | None = None,
         overrides: Params | list[str] | tuple[str, ...] | None = None,
     ) -> "Config":
+        if cfg is not None:
```
klemen1999, line 866: With this change [this](https://github.com/luxonis/luxonis-train/blob/b0d757c35a720bf6842b7ea293c1f9b025999358/luxonis_train/__main__.py#L447) is probably not needed anymore?

kozlov721: It can still be called independendlty from elsewhere.

klemen1999: Yeah but in the flow now we call upgrade_config once and then it gets called again in the next line implicitly because we init LuxonisModel, that gets config and that upgrade config. Or am I missing something?

kozlov721: Oh I didn't realize there is a link in the initial comment. Yes that's no longer necessary

PR #370, opened 2026-04-10
Fix: Checkpoint saving and loading

kozlov721:
## Purpose

Fixes a few issues with checkpoint saving and loading:

- The `UploadCheckpoint.on_save_checkpoint` method is called before the checkpoint is modified by the `LuxonisLightningModule.on_save_checkpoint`
  - Causes the uploaded checkpoint to lack the extra checkpoint fields
- Inconsistent loading of dataset metadata from a checkpoint  

## Specification

- Fixed adding our extra keys to the checkpoint in `UploadCheckpoint`
- Added `save_checkpoint` and `get_checkpoint` methods to `LuxonisModel` for easier testing
- Moved config restoration from checkpoint from `__main__` to `LuxonisModel`
- Renamed `DebugLoader` to `DummyLoader` and `debug_mode` in `LuxonisModel` to `allow_empty_dataset`
- Added `class_names` parameter to `DummyLoader`
- Renamed `debug_mode` to `allow_empty_dataset` in `LuxonisModel`

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

- Added new checkpoint tests

### DIFF
--- a/luxonis_train/__main__.py
+++ b/luxonis_train/__main__.py
@@ -16,7 +16,6 @@ from luxonis_train.upgrade import upgrade_config, upgrade_installation
 if TYPE_CHECKING:
     import numpy as np
 
-    from luxonis_train import LuxonisModel
 
 app = App(
     help="Luxonis Train CLI",
@@ -57,6 +56,8 @@ def train(
         suppresses some exceptions to allow training without a fully
         defined model.
     """
+    from luxonis_train import LuxonisModel
+
     LuxonisModel(config, opts, weights=weights, debug_mode=debug).train(
         weights=weights
     )
@@ -71,6 +72,8 @@ def tune(opts: list[str] | None = None, /, *, config: str | None = None):
     @type opts: list[str]
     @param opts: A list of optional CLI overrides of the config file.
     """
+    from luxonis_train import LuxonisModel
+
     LuxonisModel(config, opts).tune()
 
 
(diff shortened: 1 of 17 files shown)


luxonis_train/__main__.py
```diff
@@ -42,48 +46,23 @@ def create_model(
     config: PathType | Params | None,
     opts: list[str] | None = None,
     weights: PathType | None = None,
-    debug_mode: bool = False,
-    load_dataset_metadata: bool = True,
+    allow_empty_dataset: bool = False,
```
klemen1999, line 49: I don't see any description anywhere what this flag is for/what it does. We should add this info somewhere

klemen1999: Nevermind, actually I see it mentioned in the docstring of `LuxonisModel`. 
Though should we add some note in the README as well and should we unify its naming? Currently we have `--debug` in the CLI and `allow_empty_dataset` in python API which is a bit strange IMO

kozlov721: Yeah I kept it for `train`, `tune`, and `test` in the CLI as `--debug` since here this would actually serve for debugging mostly.

klemen1999: Ok then lets just add some docs in README  about this flag (what is its purpose, how can it be used, what is the default depending on the commans,...) ao its clear

kozlov721: I added some more info the the docs and README

klemen1999 (approved):

PR #375, opened 2026-04-16
Predictions written in batches when save_dir is set instead of buffer in memory

dtronmans:
## Purpose
- `luxonis_train infer --weights ... --config ... --save_dir` was only writing the images after all the predictions are made
- This takes a lot of space in the memory especially if the val set consists of many batches
- In this PR the only behavior that is modified is when `save_dir` is set: write the images after each predicted batch
- Tested locally

### DIFF
--- a/luxonis_train/core/utils/infer_utils.py
+++ b/luxonis_train/core/utils/infer_utils.py
@@ -1,6 +1,6 @@
 from collections import defaultdict
 from collections.abc import Iterable
-from contextlib import suppress
+from contextlib import nullcontext, suppress
 from pathlib import Path
 from typing import Any, Literal, cast
 
@@ -8,7 +8,6 @@ import cv2
 import numpy as np
 import torch
 import torch.utils.data as torch_data
-from lightning.pytorch.callbacks import BasePredictionWriter
 from loguru import logger
 from luxonis_ml.data import DatasetIterator, LuxonisDataset
 from luxonis_ml.typing import PathType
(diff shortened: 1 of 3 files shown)


luxonis_train/core/utils/infer_utils.py
```diff
@@ -145,14 +145,31 @@ def infer_from_loader(
...
+        # Save outputs as each batch is predicted so we do not keep the full
+        # prediction list in memory before writing visualizations to disk.
+        save_dir = Path(save_dir)
+        lt_module = model.lightning_module.eval()
+        counter = Counter()
+        trainer = cast(Any, model.pl_trainer)
+
+        for batch in loader:
+            with torch.inference_mode(), _get_predict_context(trainer):
+                outputs = lt_module.predict_step(batch)
```
klemen1999: This has potential issues with GPU only machines because it directly uses the predict_step of the model and not the trainer.predict(). So if model is on CUDA then output is on CUDA and saving this tensor will error out.
There is a nicer way to achieve what we want through a temporary callback. I'll sync with you offline.

klemen1999 (approved):
LGTM
I would just add a small test around the adding/removing the callback. So have an instance of LuxonisModel, call `infer` on it, check that callback is not there anymore, maybe call `.test() ` and re-check and callback is not there and then call `.infer()` again on the same instance and we can test that the images are saved to same paths (ie. tha the `counter` is reset). 
Feel free to re-design this test as you want, but the idea is just to test that callback doesn't "leak out"

dtronmans:
> LGTM I would just add a small test around the adding/removing the callback. So have an instance of LuxonisModel, call `infer` on it, check that callback is not there anymore, maybe call `.test() ` and re-check and callback is not there and then call `.infer()` again on the same instance and we can test that the images are saved to same paths (ie. tha the `counter` is reset). Feel free to re-design this test as you want, but the idea is just to test that callback doesn't "leak out"

Done, just a few notes regarding what is being asserted:

- `infer_from_loader` receives the `MockTrainer` through the `model.pl_trainer`, so if it's called with `save_dir` then we're trying to assert that it should only attach the callback throuhg the duration of the infer call.
- If I call `trainer.predict()` directly afterwards then nothing should be saved because the callback should have been removed at the end of the previous `infer_from_loader` call
- If I call `infer_from_loader` again then the image is saved under the same name as the image that was saved by the previous `infer_from_loader` call, meaning that the callback state was reset (so a new callback from a fresh state was added to the trainer during the call)

klemen1999 (approved):
LGTM, feel free to merge after tests pass

PR #377, opened 2026-04-24
Skip passing normalization if it is not active in the config

dtronmans:
## Purpose
Resulting ONNX config:
resulting_output_onnx

Resulting RVC4 config:
resulting_output_rvc4

`None` for mean and scale is properly handled in `modelconverter`:

- https://github.com/luxonis/modelconverter/blob/main/modelconverter/utils/onnx_tools.py#L139
- https://github.com/luxonis/modelconverter/blob/main/modelconverter/utils/onnx_tools.py#L241
- https://github.com/luxonis/modelconverter/blob/main/modelconverter/utils/onnx_tools.py#L212

Another option would be to manually pass the identity (so [0, 0, 0] [1, 1, 1] for RGB and [0] [1] for grayscale) to the ONNX archive if normalization is not active, instead of passing `None` and letting it be handled downstream by `modelconverter`

### DIFF
--- a/luxonis_train/core/utils/export_utils.py
+++ b/luxonis_train/core/utils/export_utils.py
@@ -63,6 +63,9 @@ def get_preprocessing(
 ) -> tuple[
     list[float] | None, list[float] | None, Literal["RGB", "BGR", "GRAY"]
 ]:
+    if not cfg.normalize.active:
+        return None, None, cfg.color_space
+
     def _get_norm_param(key: Literal["mean", "std"]) -> list[float] | None:
         params = cfg.normalize.params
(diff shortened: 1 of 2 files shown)


klemen1999 (approved):
LGTM. And yes, I think is fine to have `null` values and let other tools (modelconverter) handle this correctly since we do allow None values for this parameters [in the schema](https://github.com/luxonis/luxonis-ml/blob/bd5cb4dc275043f19b444276c0e2200abcc364da/luxonis_ml/nn_archive/config_building_blocks/base_models/input.py#L34).

kozlov721 (approved):

PR #378, opened 2026-04-27
--list-augmentations for inspect command

dtronmans:
## Purpose
Relies on methods defined in the corresponding `luxonis-ml` PR: https://github.com/luxonis/luxonis-ml/pull/408/changes

Usage: `luxonis_train inspect --config <config-path> --list-augmentations`

Example (with `use_for_resizing`):

lt-Use_for_resizing

### DIFF
--- a/luxonis_train/__main__.py
+++ b/luxonis_train/__main__.py
@@ -121,12 +121,33 @@ def _yield_visualizations(
     size_multiplier: Annotated[
         float, Parameter(["--size_multiplier", "-s"])
     ] = 1.0,
+    list_augmentations: bool = False,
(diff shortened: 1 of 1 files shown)


kozlov721 (approved):

klemen1999 (commented):
LGTM but this needs also luxonis-ml version bump right? In that case let's wait with merge until next LuxonisML version is released

dtronmans:
> LGTM but this needs also luxonis-ml version bump right? In that case let's wait with merge until next LuxonisML version is released

True, this only works here because the workflows point to `luxonis-ml`'s `main` branch instead of a release.

PR #384, opened 2026-05-18
`resume_training` false with weights set loads weights but resets optimizer/scheduler, `strict_weights_loading` argument

dtronmans:
## Purpose
- New model-side config item `strict_weights_loading` that fails immediately if any part of the model state dict is different.
- If `resume_training` is set to False and weights are passed, then the weights are still loaded but the optimizer/scheduler states are taken from the config. Previously this combination of `resume_training` and weights being passed would create a fresh model instead of loading the weights.
- If `resume_training` is set to True, then the behavior is unchanged (from #326)
- Small bug fix: passing CLI weights would warn `Weights provided on the command line, but config weights are set. " "Ignoring weights provided in config or during LuxonisModel initialization` even if the config weights were not set, this has been patched. I had to add `self._weights_provided_during_init` and `self._weights_provided_in_config` attributes because one level lower (in `resolve_weights`) it has no indication of which weights came from the config or the CLI.

## Testing & Validation
- Training a model from the [DetectionModel](https://github.com/luxonis/luxonis-train/blob/main/configs/detection_light_model.yaml) config then creating a new checkpoint from this model with an extra key `repvgg_encoder.scale_layer.conv.weight.__unexpected__` crashed quickly on `train` and `infer` if `strict_weights_loading` is set to True:
strict_true

- If `strict_weights_loading` is set to False (default), training and inference continue even with mis-matching state dicts:
strict_false

- `resume_training` set to False and weights are passed: starts from epoch 0 again with only the checkpoint loaded but not the surrounding (optimizer/scheduler/epoch count etc) state.
resume_training_false_warn
resume_training_false

## AI Usage

### DIFF
--- a/configs/defaults.yaml
+++ b/configs/defaults.yaml
@@ -2,6 +2,7 @@ rich_logging: true
 
 model:
   name: model
+  strict_weights_loading: false
   nodes: []
   outputs: []
 
--- a/configs/defaults.yaml
+++ b/configs/defaults.yaml
@@ -2,7 +2,6 @@ rich_logging: true
 
 model:
   name: model
-  strict_weights_loading: false
   nodes: []
   outputs: []
 
@@ -58,6 +57,7 @@ trainer:
   epochs: 100
   overfit_batches: 0
   resume_training: false
+  strict_weights_loading: false
 
   n_workers: 4
   validation_interval: 5
--- a/luxonis_train/lightning/luxonis_lightning.py
+++ b/luxonis_train/lightning/luxonis_lightning.py
@@ -25,7 +25,10 @@ from luxonis_train.config import Config
 from luxonis_train.nodes import BaseNode
 from luxonis_train.typing import Labels, Packet
 from luxonis_train.utils import DatasetMetadata, LuxonisTrackerPL
-from luxonis_train.utils.checkpoint import filter_checkpoint_state_dict
+from luxonis_train.utils.checkpoint import (
+    CHECKPOINT_FILTERED_STATE_DICT_PATTERN,
+    filter_checkpoint_state_dict,
+)
 
 from .luxonis_output import LuxonisOutput
 from .utils import (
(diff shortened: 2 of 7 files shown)


configs/defaults.yaml
```diff
@@ -2,6 +2,7 @@ rich_logging: true
 
 model:
   name: model
+  strict_weights_loading: false
```
klemen1999: This new config parameter should be under `trainer` section

dtronmans: Addressed in [604db1f](https://github.com/luxonis/luxonis-train/pull/384/commits/604db1fed128d0b2218b12a52d5fd63daa0a0640) and tested the different combinations of `resume_training` and `strict_weights_loading` again manually

luxonis_train/lightning/luxonis_lightning.py
```diff
@@ -167,9 +170,41 @@ def load_state_dict(
 
         In case resume_training is active, allow loading in a non-strict
         manner to allow loss, visualizer and metric nodes to be absent.
+        When strict weight loading is enabled, only those filtered
+        attached-module keys may be missing or unexpected.
         """
         if self.cfg.trainer.resume_training:
-            return super().load_state_dict(state_dict, strict=False)
+            incompatible = super().load_state_dict(state_dict, strict=False)
```
klemen1999: Problem with having this on the top of the function is that if for some reason some attached module (which we below allow it to be different and ignore it) has changed shape then this `load_state_dict(state_dict, strict=False)` will still raise RuntimeError due to shape mismatch. Meaning that a module that we were going to ignore anyway would crash before we have a chance to ignore it

luxonis_train/lightning/luxonis_lightning.py
```diff
@@ -167,9 +170,41 @@ def load_state_dict(
...
+                        if missing_keys
+                        else ""
+                    )
+                    + (
+                        f"\tUnexpected key(s): {', '.join(unexpected_keys)}.\n"
+                        if unexpected_keys
+                        else ""
+                    )
+                )
+            return _IncompatibleKeys([], [])
```
klemen1999, line 214: Safer implementation of this would be something like this (+it reuses already implemented  `filter_checkpoint_state_dict` function):
```
def load_state_dict(
    self, state_dict: Mapping[str, Tensor], strict: bool = True
) -> _IncompatibleKeys:
    if self.cfg.trainer.resume_training:
        filtered_state_dict = (
            filter_checkpoint_state_dict(state_dict)
            if self.cfg.trainer.strict_weights_loading
            else state_dict
        )

        incompatible = super().load_state_dict(
            filtered_state_dict, strict=False
        )
        if not self.cfg.trainer.strict_weights_loading:
            return incompatible

        if incompatible.missing_keys or incompatible.unexpected_keys:
            raise RuntimeError(
                "Error(s) in loading state_dict for "
                f"{self.__class__.__name__}:\n"
                + (
                    f"\tMissing key(s): {', '.join(incompatible.missing_keys)}.\n"
                    if incompatible.missing_keys
                    else ""
                )
                + (
                    f"\tUnexpected key(s): {', '.join(incompatible.unexpected_keys)}.\n"
                    if incompatible.unexpected_keys
                    else ""
                )
            )
        return _IncompatibleKeys([], [])

    return super().load_state_dict(state_dict, strict=strict)

```

dtronmans: True, addressed in [1d1ee8e](https://github.com/luxonis/luxonis-train/pull/384/commits/1d1ee8e4383afb6bffdf399f2c31aa02013218e7). Since this could be a sensitive feature I added three checkpoints (`original_ckpt.ckpt`, `modified_model_ckpt.ckpt` and `modified_attached_modules_ckpt.ckpt`) to GCP and added tests for:

- `strict_weights_loading` False doesn't fail if I add an `__unexpected__` key to a model state dict
- `strict_weights_loading` True fails if I add an `__unexpected__` key to a model state dict
- `strict_weights_loading` True doesn't fail if I change if an attached loss/metric/visualizer module changes shape (`modified_attached_modules_ckpt.ckpt`) because those checkpoint keys are supposed to be filtered out before loading

klemen1999 (approved):

PR #403, opened 2026-06-15
Feat/keypoint per class metrics

dtronmans:
## Purpose
- `MeanAveragePrecisionKeypoints` now accepts `class_metrics`, which computes per-class `kpt_map`, `kpt_mar` and `kpt_f1`.
- Expose the `per_class_metrics` -> `class_metrics` alias also for Tasks.INSTANCE_KEYPOINTS and Tasks.INSTANCE_SEGMENTATION_KEYPOINTS in the mAP factory
- Updated docs: added `class_metrics` to the keypoint mAP params table and describe the two ways per-class metrics can be enabled (either through predefined model or metric itself)
- Updated docs: The predefined model `KeypointDetectionModel` incorrectly showed `ObjectKeypointSimilarity` as the primary keypoint metric, in the code it was `MeanAveragePrecision` so updated this. Added `per_class_metrics` to the predefined model param.

- Extra (unrelated) change: AIMET path loading and serialization tests to hit the new `config.py` 100% coverage requirement.

## Specification
Two ways of getting per-class metrics: predefined model-level and metric-level

Predefined model:
set_config

Individual metrics:
set_metric_level

## AI Usage

Assisted-by: CODEX

## Summary by CodeRabbit

* **New Features**
  * Added per-class keypoint metrics support, enabling class-specific calculations for keypoint precision, recall, and F1 scores.

* **Documentation**
  * Updated metrics documentation with `per_class_metrics` parameter configuration options.
  * Corrected `box_format` typo in instance keypoint params table.
  * Updated keypoint detection model metrics hierarchy and parameters.

* **Tests**
  * Added test coverage for per-class keypoint metrics functionality.

### DIFF
--- a/luxonis_train/attached_modules/metrics/mean_average_precision/mean_average_precision_bbox.py
+++ b/luxonis_train/attached_modules/metrics/mean_average_precision/mean_average_precision_bbox.py
@@ -16,8 +16,10 @@ class MeanAveragePrecisionBBox(MeanAveragePrecision, BaseMetric):
         Tasks.INSTANCE_SEGMENTATION_KEYPOINTS,
     ]
 
-    def __init__(self, **kwargs):
-        super().__init__(iou_type="bbox", **kwargs)
+    def __init__(self, class_metrics: bool = False, **kwargs):
+        super().__init__(
+            iou_type="bbox", class_metrics=class_metrics, **kwargs
+        )
 
     @override
     def update(
--- a/luxonis_train/attached_modules/metrics/mean_average_precision/mean_average_precision_bbox.py
+++ b/luxonis_train/attached_modules/metrics/mean_average_precision/mean_average_precision_bbox.py
@@ -16,10 +16,8 @@ class MeanAveragePrecisionBBox(MeanAveragePrecision, BaseMetric):
         Tasks.INSTANCE_SEGMENTATION_KEYPOINTS,
     ]
 
-    def __init__(self, class_metrics: bool = False, **kwargs):
-        super().__init__(
-            iou_type="bbox", class_metrics=class_metrics, **kwargs
-        )
+    def __init__(self, **kwargs):
+        super().__init__(iou_type="bbox", **kwargs)
 
     @override
     def update(
--- a/luxonis_train/attached_modules/metrics/mean_average_precision/mean_average_precision_keypoints.py
+++ b/luxonis_train/attached_modules/metrics/mean_average_precision/mean_average_precision_keypoints.py
@@ -10,6 +10,7 @@ from typing_extensions import override
 from luxonis_train.attached_modules.metrics import BaseMetric, MetricState
 from luxonis_train.attached_modules.metrics.mean_average_precision.utils import (
     add_f1_metrics,
+    postprocess_metrics,
 )
 from luxonis_train.attached_modules.metrics.utils import (
     fix_empty_tensor,
@@ -51,6 +52,7 @@ class MeanAveragePrecisionKeypoints(BaseMetric):
         area_factor: float | None = None,
         max_dets: int = 20,
         box_format: Literal["xyxy", "xywh", "cxcywh"] = "xyxy",
+        class_metrics: bool = False,
         **kwargs,
     ):
         """
@@ -65,6 +67,9 @@ class MeanAveragePrecisionKeypoints(BaseMetric):
         @param max_dets: Maximum number of detections to be considered per image. Defaults to C{20}.
         @type box_format: Literal["xyxy", "xywh", "cxcywh"]
         @param box_format: Input bounding box format. Defaults to C{"xyxy"}.
+        @type class_metrics: bool
+        @param class_metrics: Whether to calculate per-class keypoint metrics.
+            Defaults to C{False}.
         """
         super().__init__(**kwargs)
 
@@ -74,6 +79,7 @@ class MeanAveragePrecisionKeypoints(BaseMetric):
         )
         self.max_dets = max_dets
         self.box_format = box_format
+        self.class_metrics = class_metrics
 
     @override
     def update(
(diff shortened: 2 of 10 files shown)


luxonis_train/attached_modules/metrics/mean_average_precision/mean_average_precision_keypoints.py
```diff
@@ -141,18 +147,65 @@ def compute(self) -> tuple[Tensor, dict[str, Tensor]]:
...
+            "kpt_mar_medium": stats[8],
+            "kpt_mar_large": stats[9],
+        }
+
+        if self.class_metrics:
+            metrics.update(self._compute_class_metrics())
+
+        return postprocess_metrics(
+            add_f1_metrics(metrics),
+            self.classes.inverse,
```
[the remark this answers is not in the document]
dtronmans: I understand, but wouldn't this lead to inconsistent computations between different f1 metrics? Because other keypoint-related metrics like kpt_f1_75 can also lead to nan values. So it would be a bit inconsistent to have kpt_f1_50 and kpt_f1_75 (for example) to have nan values (current behavior) but for nan to be "represented" by 0 for kpt_f1_per_class_<class_name> metrics. Let me know what you think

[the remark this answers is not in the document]
dtronmans: I think what you write makes sense, and there is another problem (the fact that NaN sub-metrics don't actually show on MLflow until they have a numeric value like 0), but changing the interpretation/semantic meaning of the f1 metrics is not in-scope for this PR, the only things I wanted to do was fix the READMEs and add per-class keypoints metrics that are consistent with the way they are currently implemented. So let's leave it for now.

luxonis_train/attached_modules/metrics/mean_average_precision/mean_average_precision_bbox.py
```diff
@@ -16,8 +16,10 @@ class MeanAveragePrecisionBBox(MeanAveragePrecision, BaseMetric):
         Tasks.INSTANCE_SEGMENTATION_KEYPOINTS,
     ]
 
-    def __init__(self, **kwargs):
-        super().__init__(iou_type="bbox", **kwargs)
+    def __init__(self, class_metrics: bool = False, **kwargs):
+        super().__init__(
+            iou_type="bbox", class_metrics=class_metrics, **kwargs
```
klemen1999: Why did we explicitly add class_metrics argument? Wasn't this working before as well since we would pass this argument through the kwargs?

tests/unittests/test_config.py
```diff
@@ -524,6 +526,52 @@ def test_export_config_validation():
         HubAIExportConfig(platform="hailo")
 
 
+def test_aimet_config_path_loading_and_serialization(
```
klemen1999, line 529: I'm fine with squeezing this into this PR, could you just explain why was it added? Is it to satisfy config.py test converage which wasn't noticed before?

luxonis_train/attached_modules/metrics/mean_average_precision/mean_average_precision_keypoints.py
```diff
@@ -141,18 +147,65 @@ def compute(self) -> tuple[Tensor, dict[str, Tensor]]:
...
+            "kpt_mar_75": stats[7],
+            "kpt_mar_medium": stats[8],
+            "kpt_mar_large": stats[9],
+        }
+
+        if self.class_metrics:
+            metrics.update(self._compute_class_metrics())
+
+        return postprocess_metrics(
+            add_f1_metrics(metrics),
```
klemen1999: We add_f1_metrics() here but the first line in postprocess_metrics already calls add_f1_metrics(). So isn't this redundant?

luxonis_train/attached_modules/metrics/mean_average_precision/utils.py
```diff
@@ -35,7 +35,10 @@ def add_f1_metrics(metrics: dict[str, Tensor]) -> dict[str, Tensor]:
 def process_class_metrics(
     metrics: dict[str, Tensor], class_names: Mapping[int, str]
 ) -> dict[str, Tensor]:
-    classes = metrics.pop("classes")
+    classes = metrics.pop("classes", None)
+    if classes is None:
```
klemen1999: Likely shouldn't be needed after the change in kpt mAP base keys

tests/unittests/test_config.py
dtronmans: Yes forgot to mention, this is because of the new 100% coverage requirement. I will add a note in the PR comment about this

luxonis_train/attached_modules/metrics/mean_average_precision/mean_average_precision_keypoints.py
dtronmans: True good catch, I removed add_f1_metrics when they are passed to postprocess_metrics in [88758aa](https://github.com/luxonis/luxonis-train/pull/403/commits/88758aa665add61610bcf1dc4967765417aaf307)

luxonis_train/attached_modules/metrics/mean_average_precision/utils.py
dtronmans: Addressed in [88758aa](https://github.com/luxonis/luxonis-train/pull/403/commits/88758aa665add61610bcf1dc4967765417aaf307)

luxonis_train/attached_modules/metrics/mean_average_precision/mean_average_precision_bbox.py
dtronmans: True, addressed in [88758aa](https://github.com/luxonis/luxonis-train/pull/403/commits/88758aa665add61610bcf1dc4967765417aaf307)

klemen1999 (approved):

PR #406, opened 2026-06-18
PTQ and QAT fixes and QoL improvements

dtronmans:
## Purpose
List of changes by importance/severity:
- QAT after PTQ could crash because Lightning evaluation runs used `inference_mode=True`, which caused tensors from the PTQ/test pass to be reused later in QAT under autograd. Fixed by running the quantization evaluation passes with a dedicated trainer configured with `inference_mode=False`
- It's possible that post-QAT batch norm folding fails when AIMET tries to rebuild a `ConnectedGraph`. It was previously crashing, now just log a warning that the step is skipped.
- Since the `val` loader is passed to PTQ, this means that it was previously limited by the trainer config's `n_validation_batches` parameters, which could be an implication the user is unaware of since `n_validation_batches` is a training-related parameter. So `n_validation_batches` no longer impacts PTQ, and instead there is a new `exporter.aimet.max_calibration_images` config item that limits the number of calibration images.
- QAT's loading bar was previously just one bar, with each step corresponding to an epoch. Replaced the granularity so that there is one bar per epoch, and each step is a batch instead.
- Log the output dir of QAT the same way the end of the `train` command does.

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation
**QAT crash because of inference tensors**

Previous QAT crash because of `inference_mode=True`:
inference-mode-tensors

**Batch norm folding crash**

Previous error with post-QAT batch norm folding:
previous_unsafegrapherror

Now an emitted warning instead:
bnfoldingerror

**Progress bar**

After fixing, QAT runs but with one progress bar, which has epoch-level steps:
qat_epoch_steps

Changed to different epochs with batch-level steps:
new_qat_granularity

**Misc**

Using the new config item `max_calibration_images`:
aimet_subset_images

Log AIMET output dir:
log_output_folder

## AI Usage
Assisted-by: CODEX

## Release Notes

* **New Features**
  * Added `max_calibration_images` configuration parameter to optionally limit the number of calibration images used in AIMET quantization (independent of validation batch settings).

* **Bug Fixes**
  * Improved quantization calibration accuracy during post-training quantization.

* **Documentation**
  * Updated configuration documentation with new calibration image limit parameter and usage details.


klemen1999 (approved):
Didn't practically test but going through the code myself and with agents it looks good to me

PR #411, opened 2026-06-25
Correct ONNX NNArchive output names after QAT process

dtronmans:
## Purpose
In the quantization path, output names were being computed before export mode was enabled, which caused the exported ONNX to keep intermediate /<model>/... names instead (like ../EfficientBBoxHead/..) of the expected output{i}_yolov6r2 names. The fix computes output names after set_export_mode(True) to align AIMET exports with the normal ONNX export path and the generated NN archive metadata.

Correct ONNX NNArchive output names after the fix:
after_correct

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

## AI Usage
Assisted-by: CODEX

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ONNX export during quantization so output tensor names are preserved and written into the exported model.
  * Strengthened handling of model inputs during quantization, with clearer enforcement of the single-input limitation.
  * Added safer model file rewriting when external ONNX data files are present.

### DIFF
--- a/luxonis_train/core/utils/export_utils.py
+++ b/luxonis_train/core/utils/export_utils.py
@@ -1,4 +1,3 @@
-import json
 import os
 import shutil
 from collections.abc import Generator
@@ -64,41 +63,13 @@ def rename_onnx_outputs(onnx_path: PathType, output_names: list[str]) -> None:
 
     onnx_path = Path(onnx_path)
     model = onnx.load(str(onnx_path))
-    actual_outputs = []
-    for output in model.graph.output:
-        shape = [dim.dim_value for dim in output.type.tensor_type.shape.dim]
-        actual_outputs.append(
-            {
-                "name": output.name,
-                "shape": shape,
-                "dtype": onnx.TensorProto.DataType.Name(
-                    output.type.tensor_type.elem_type
-                ),
-            }
-        )
-
-    logger.info(f"Requested ONNX output names: {output_names}")
-    logger.info(f"Actual exported ONNX outputs: {actual_outputs}")
-
-    debug_path = onnx_path.with_suffix(".output_debug.json")
-    debug_path.write_text(
-        json.dumps(
-            {
-                "requested_output_names": output_names,
-                "actual_outputs": actual_outputs,
-            },
-            indent=2,
-        )
-    )
-    logger.info(f"Wrote ONNX output debug info to {debug_path}")
 
     if len(model.graph.output) != len(output_names):
-        logger.error(
+        raise ValueError(
             "Number of requested output names does not match the number "
             f"of ONNX graph outputs: {len(output_names)} != "
-            f"{len(model.graph.output)}. Skipping output rename."
+            f"{len(model.graph.output)}."
         )
-        return
 
     for output, new_name in zip(model.graph.output, output_names, strict=True):
         output.name = new_name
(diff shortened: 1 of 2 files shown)


luxonis_train/core/utils/export_utils.py
```diff
@@ -58,6 +58,38 @@ def try_onnx_simplify(onnx_path: PathType) -> None:
...
+
+    if len(model.graph.output) != len(output_names):
+        raise ValueError(
+            "Number of requested output names does not match the number "
+            f"of ONNX graph outputs: {len(output_names)} != "
+            f"{len(model.graph.output)}."
+        )
+
+    for output, new_name in zip(model.graph.output, output_names, strict=True):
+        output.name = new_name
```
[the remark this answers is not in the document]
dtronmans: You're right, I'm only renaming the graph’s declared outputs, not the actual tensor values produced by the last nodes. Addressed in [64b3569](https://github.com/luxonis/luxonis-train/pull/411/commits/64b3569827ae94a1023e929c0d648e583d2dbf1c)

luxonis_train/core/utils/export_utils.py
```diff
@@ -58,6 +58,53 @@ def try_onnx_simplify(onnx_path: PathType) -> None:
     logger.info(f"ONNX model saved to {onnx_path}")
 
 
+def rename_onnx_outputs(onnx_path: PathType, output_names: list[str]) -> None:
```
klemen1999, line 61: Can we get away with something like this?
```python
def rename_onnx_outputs(onnx_path: PathType, output_names: list[str]) -> None:
    import onnx

    onnx_path = Path(onnx_path)
    model = onnx.load(str(onnx_path))

    if len(model.graph.output) != len(output_names):
        raise ValueError(
            "Number of requested output names does not match the number "
            f"of ONNX graph outputs: {len(output_names)} != "
            f"{len(model.graph.output)}."
        )

    for output, new_name in zip(model.graph.output, output_names, strict=True):
        output.name = new_name

    external_data_path = onnx_path.with_name(f"{onnx_path.name}.data")
    if external_data_path.exists():
        external_data_path.unlink()
        onnx.save(
            model,
            str(onnx_path),
            save_as_external_data=True,
            all_tensors_to_one_file=True,
            location=external_data_path.name,
            size_threshold=0,
            convert_attribute=False,
        )
    else:
        onnx.save(model, str(onnx_path))

    onnx.checker.check_model(str(onnx_path))

```

Basically only touching the node.name property of the ONNX node instead of touching the internal wiring by changing the `.output`.

dtronmans: This is what I had before [64b3569](https://github.com/luxonis/luxonis-train/pull/411/commits/64b3569827ae94a1023e929c0d648e583d2dbf1c), changing the `graph.output` was implemented after a [comment by CodeRabbit](https://github.com/luxonis/luxonis-train/pull/411#discussion_r3474690253). Functionally both (touching only `node.name` or both `node.name` and `graph.output`) have led to normal on-device results when I tested exporting, converting then running on device, so I can revert it.

klemen1999: Ahh I see, seems like we have a clash in what LLMs think is the correct answer here :sweat_smile: 
Based on [this](https://github.com/onnx/onnx/issues/2052#issuecomment-495384662) from onnx official repo it looks like CodeRabbit is right and we should stick with what we have. Generally you'd also want to search and change .inputs on the nodes that have these renamed nodes as inputs but here these should be leaf nodes anyway so this shouldn't matter. 
So TLDR should be fine to leave as is right now

klemen1999 (approved):

PR #420, opened 2026-07-22
Packaged configuration and versioning of predefined models

kozlov721:
## Purpose

1. CLI shortcuts for packaged predefined models. Running a bundled example used to require `luxonis_train train --config configs/detection_light_model.yaml`, a long path that also did not work for pip-installed users because `configs/` was never shipped in the wheel. This adds `--model <name>` and optional `--variant` arguments to every command, so `luxonis_train train --model detection --variant light` is equivalent to the long form. The configuration YAMLs now ship inside the wheel, and the Python API supports the same selection through `LuxonisModel(model=..., variant=...)`.
2. Versioning for predefined-model classes. Without a version marker, a future breaking change to (*e.g.*) `DetectionModel`'s default backbone would silently break existing checkpoints with a cryptic PyTorch shape error. This introduces a version field on the `predefined_model` configuration block, restructures the classes on disk as `predefined_models/<family>/v<N>/model.py`, and changes the registry keys to `Family:vN` (*e.g.* `DetectionModel:v1`). Old configurations keep working (default version: `"latest"`). A mismatch between checkpoint-time and load-time resolution now produces a clear warning telling the user which version to pin to reproduce the old architecture.
   - `luxonis_train train --model detection:v2 --variant heavy`
3. Extra helper CLI commands, `list-models` and `info`.
   - `luxonis_train info --model detection --variant heavy`

### CLI predefined-model shortcuts

- New module `luxonis_train/config/predefined.py` with utilities to resolve model names and variants.
- `luxonis_train/configs/__init__.py` (new, empty) and a `pyproject.toml` package entry.
- `luxonis_train/__main__.py`: every command gains `--model` and `--variant` keyword-only parameters, and the Management group gains a `list-models` command.
- `--model` and `--config` are mutually exclusive. `--variant` requires `--model`. Malformed inputs surface clear errors before Lightning is imported.
- Added the embeddings model to the predefined models.

### Predefined-model versioning

- Files moved: each of the eight predefined models is now at `luxonis_train/config/predefined_models/<family>/v1/model.py`. `predefined_models/__init__.py` re-exports the classes so existing imports keep working.
- `BasePredefinedModel` gains `_VERSION: int = 1`. New iterations of a model (*e.g.* a future V2) can be added at `<family>/v2/model.py` with `_VERSION = 2`.
- Registry keys are now `Family:vN` (`DetectionModel:v1`, `ClassificationModel:v1`, ...).
- New module `luxonis_train/config/predefined_versions.py`.
- `ModelConfig.validate_predefined_model` now resolves through `resolve_predefined_class` and logs the resolved key when it differs from the requested name.
- `LuxonisLightningModule.load_checkpoint` calls the mismatch warning using the configuration stored in each checkpoint.
- CLI: `--model detection:v1` and `--model detection:latest` are supported.
- `list-models` shows both variants and available versions for each model.

### Backward compatibility

Existing predefined-model configuration contents load unchanged. A bare `predefined_model.name: DetectionModel` with no version field resolves to `DetectionModel:v1`. Existing checkpoints without `predefined_model.version` metadata are accepted without a mismatch warning. Public imports such as `from luxonis_train.config.predefined_models import DetectionModel` continue to work.

Not applicable. There are no new runtime dependencies, service-level impacts, or breaking changes to the configuration schema. The version field is additive and has a safe default.

## Deployment Plan

Not applicable. This is a library release with no rollout or rollback concerns.

## AI Usage

Assisted-by: Claude:claude-opus-4-7

The author is taking responsibility for the contribution: YES

## Summary
* **New Features**
  * Select packaged predefined models and variants through the CLI (`--model`/`--variant`) across train/tune/inspect/test/infer/annotate/export/archive/convert/quantize.
  * Added `list-models` to browse available models, variants, and versions, plus `info` to view model details.
  * Added `EmbeddingsModel`, used by the shipped embeddings configuration.
* **Bug Fixes**
  * Save predefined-model metadata in checkpoints and warn on mismatches when resuming.
  * Enforced CLI rules: `--variant` requires `--model`, and `--config` cannot be combined with `--model`.
* **Documentation**
  * Updated README and configuration tips to prefer `--model` workflows with version pinning.
* **Tests**
  * Expanded packaged-configuration/version resolution and checkpoint-warning test coverage.


luxonis_train/lightning/luxonis_lightning.py
```diff
@@ -50,6 +51,27 @@
 )
 
 
+def _checkpoint_predefined_model(cfg: Config) -> dict[str, Any] | None:
```
klemen1999: Nitpick: I'd put this below class definition

luxonis_train/lightning/luxonis_lightning.py
```diff
@@ -50,6 +51,27 @@
...
+    dumped = predefined_model.model_dump()
+    if dumped.get("version") == "latest":
+        from luxonis_train.config.predefined_versions import (
+            resolve_predefined_class,
+        )
+
+        # Read the version off the resolved class rather than parsing it
+        # out of the registry key: a model registered under a plain key
+        # has no `:vN` suffix to parse, and would keep `"latest"` here -
+        # which is exactly the ambiguity the pin exists to remove.
```
klemen1999: Feels like a comment that was a result of some very specific debug Agent session. Let's make comments generally applicable, not so focused in wording on a specific debug session.

luxonis_train/lightning/luxonis_lightning.py
```diff
@@ -728,6 +761,25 @@ def _check_valid_epoch_counts(self, ckpt_config: dict) -> None:
...
+    def _warn_on_predefined_model_mismatch(
+        self, ckpt_predefined_model: Any | None
+    ) -> None:
+        """Warn when the checkpoint was trained with a different
+        predefined-model version than the config resolves to.
+
+        The pin is read from the checkpoint's top-level
+        `predefined_model` key only. It is deliberately not looked for
+        under `config.model.predefined_model`: that field is
+        `Field(exclude=True)`, so no checkpoint has ever contained it.
```
klemen1999: Docstring here also feels very AI generated and over focused on a specific issue that was probably debugged

README.md
```diff
@@ -358,15 +381,17 @@ Once your configuration file and dataset are ready, start the training process.
 **CLI:**
 
 ```bash
-luxonis_train train --config configs/detection_light_model.yaml
+luxonis_train train --model detection --variant light
```
klemen1999, line 434: For practical usecase I don't think just this command is enough because user probably didn't adjust the `luxonis_train/configs/detection_light_model.yaml` config and was rather adjusting their own `config.yaml`. So just running this will not result in their custom training. This should be clear and practically we should probably at least show how to pass dataset_name override since this is usually the most important part to change when training with a predefined config

klemen1999 (commented):
Generally LGTM but maybe we can make the README a bit more clear around what config is used when predefined models are selected and maybe show an example of what an actual practical command would look like (e.g. dataset name override)

klemen1999 (approved):

PR #421, opened 2026-07-22
Migration to `uv`

kozlov721:
## Purpose

Move dependency and build management to `uv`, so local development, CI and publishing all
resolve from one locked environment instead of hand-maintained `requirements*.txt` files that
drift from `pyproject.toml`.

Additionally, `uv` speeds up the execution of each matrix in our CI by a ~2 minutes compounding to ~10 saved minutes on the heavy runners per CI run. 

## Specification

- Runtime dependencies, the `aimet` extra and the `dev`/`docs`/`config-test` groups now live in
  `pyproject.toml`. The version moved to `luxonis_train/_version.py` so setuptools can read it
  statically.
- Added `.python-version` and `uv.lock`.
- `torch`/`torchvision` resolve from the cu130 index via `[tool.uv.sources]` and an
  `explicit = true` index.
- `requirements*.txt` stay, but they are now generated pip-compatibility exports produced by
  `scripts/export_requirements.sh` and refreshed by a pre-commit hook. Don't edit them by hand.
- CI, publishing and the contributor docs use `uv sync`, `uv run` and `uv build`.
- Dropped `--extra-index-url` from the documented `pip install luxonis-train[aimet]` command
  (README and the `check_aimet_available()` message). The `torch==2.11.0` on PyPI already
  bundles CUDA 13 — the same `nvidia-cuda-runtime` 13.0.96 and `nvidia-cudnn-cu13` 9.19.0.56
  the cu130 index serves — so the flag was redundant. It never applied to PyPI installs anyway,
  since `[tool.uv]` lives in `pyproject.toml` and isn't part of the built wheel.

`aimet` stays an extra because it's a user-facing feature people install as
`luxonis-train[aimet]`, and only extras end up in published metadata. `dev`, `docs` and
`config-test` are PEP 735 groups: they aren't published, so changing them can't break a
consumer, and `uv sync --only-group config-test` installs them *without* the project — which is
what lets the config-test job build an environment containing `luxonis_train` and none of its
runtime dependencies.

### Development commands

| Before | Now |
|---|---|
| `pip install -e '.[dev]'` | `uv sync` |
| `pip install -e '.[dev,aimet]' --extra-index-url https://download.pytorch.org/whl/cu130` | `uv sync --extra aimet` |
| `pre-commit install` | `uv run prek install` |
| `pre-commit run --all-files` | `uv run prek run --all-files` |
| `pytest --cov=luxonis_train --cov-report=html` | `uv run pytest --cov=luxonis_train --cov-report=html` |
| `pyright --warnings --level warning --pythonversion 3.10 luxonis_train` | `uv run pyright --warnings --project pyproject.toml` |
| `pydoctor --docformat=epytext luxonis_train` | `uv run --only-group docs pydoctor --docformat=epytext luxonis_train` |
| `python -m build` | `uv build` |
| editing `requirements.txt` by hand | edit `pyproject.toml` (or `uv add …`), then `scripts/export_requirements.sh` |

No `source .venv/bin/activate` step — `uv run` syncs `.venv` and runs in it. The
one case where that bites is a hand-installed git checkout of `luxonis-ml`, which
a sync would revert; use `uv run --no-sync <cmd>` there, as CI does.

### `opencv-python-headless`

`albumentations`, `albucore` and `roboflow` require `opencv-python-headless`, while `luxonis-ml`
and `grad-cam` require `opencv-python`. Both ship the same top-level `cv2`, so keeping both in
the lock left `cv2` as a mix of 4.10.0.84 and 4.14.0.94 decided by install order — and `uv sync`
(parallel) doesn't order things the way `pip install -r requirements.txt` (sequential) does.

uv has no "drop this dependency" switch, so the headless build is overridden to a marker that is
false on every platform:

```toml
[tool.uv]
override-dependencies = ["opencv-python-headless; sys_platform == 'never'"]
```

`opencv-python` satisfies the import for every consumer. The export script also passes
`--no-emit-package opencv-python-headless` so the generated files don't carry a dead entry.

### `prek` instead of `pre-commit`

`.pre-commit-config.yaml` is unchanged — [`prek`](https://github.com/j178/prek) reads the same
file and the same upstream hook repositories, so this is a drop-in swap rather than a
re-tooling; console output, `SKIP=` and `--show-diff-on-failure` behave the same.

It's about twice as fast on a cold cache (11.9 s vs 23.9 s), which is what CI pays on every run,
since `setup-uv`'s cache covers `UV_CACHE_DIR` and not the hook environments. It also drops six
packages from the dev resolution (297 → 291, `prek` being a single binary rather than a Python
package pulling in `virtualenv` and friends) and removes the `pre-commit>=3.2.1,<4.0.0` cap that
`main` carries with the note *"Causes issues with `filelock` dependency on luxonis-ml"* —
`filelock` arrived through `virtualenv`, which was a `pre-commit` dependency.

- `pip install -e '.[dev]'` no longer works. `dev` is a group now, not a published extra, and
  pip treats an unknown extra as a warning rather than an error — so that command exits 0 having
  installed no dev tooling. Use `uv sync`. External scripts and sibling-repo pipelines using the
  old invocation need updating.
- `requires-python` is now `>=3.10,<3.13`, matching the classifiers we already advertised but
  narrower than the previous `>=3.10`.
- `requirements.txt` is fully hash-pinned and carries `--extra-index-url` for the cu130 index, so
  pip users get the CUDA 13 torch build, and the file can't be combined with unpinned or editable
  requirements in a single command (pip enters `--require-hashes` mode).
- `requirements-dev.txt` and `requirements-aimet.txt` changed meaning: they used to be additive
  overlays (`requirements-aimet.txt` was literally three lines) and are now standalone full
  exports of base + group.
- `opencv-python-headless` is no longer installed.
- Contributors have to re-install the git hook once after pulling — `uv run prek install -f` —
  since the existing `.git/hooks/pre-commit` shim points at `pre-commit`, which is no longer in
  the dev group.

Known limitation: the `aimet` extra pins `torch==2.11`/`torchvision~=0.26`, and uv resolves a
single `torch` for the whole lock, so those are effectively the pins for every install — a plain
`uv sync` contributor can't get torch 2.9/2.10. `[tool.uv] conflicts` doesn't express this
cleanly, since CI legitimately uses `--extra aimet --group dev` together. Worth revisiting if
`aimet-torch` ever needs a torch outside the declared base range, as that would make `uv lock`
fail for all dependency changes.

## Deployment Plan

None / not applicable — packaging and CI only, no runtime services affected.

## Testing & Validation

- Reproduced the `config-test` job locally: 64 passed, 100% coverage, with `cv2`, `polars`,
  `torch` and `albumentations` confirmed absent from the environment. The group is deliberately
  `luxonis-ml[utils]` and not `[data,utils]`, so the job keeps proving that
  `luxonis_train.config` imports without luxonis-train's runtime dependencies.
- Checked the headless exclusion in a scratch project pinning `albumentations` + `roboflow` +
  `opencv-python`: headless resolves to not installed, `cv2` is 4.14.0, and `albumentations`
  imports fine against regular `opencv-python`.
- The hash-pinned visualizer tests (`test_classification_visualizer`, `test_ocr_visualizer`)
  pass against that `cv2`.
- `prek run --all-files` is green under the exact CI invocation, including the local `system`
  hook, `prettier`, `taplo` and the Python hooks. `SKIP=` still suppresses a named hook, and
  `export-requirements` is idempotent — a second run produces no diff.
- `uv lock --check` clean, all workflow YAML parses.

## AI Usage

Assisted-by: Claude:claude-opus-5

## Summary by CodeRabbit

* **New Features**
  * Updated Python support to `>=3.10,<3.13`.
  * Added structured dependency groups (`dev`, `docs`, `config-test`) and an `aimet` extra with CUDA-aware Torch/TorchVision handling.
* **Documentation**
  * Updated developer guidance to use `uv` for syncing, pre-commit, docs, type checking, and running tests.
* **Tests**
  * CI now runs type checks and test suites via `uv`, keeping existing coverage/JUnit behaviors.
* **Chores**
  * Migrated CI and publishing to `uv`, regenerated pinned requirements, added automated requirements export, refreshed pre-commit, and bumped version to **0.4.6**.


pyproject.toml
```diff
@@ -1,14 +1,41 @@
 [project]
 name = "luxonis-train"
+version = "0.4.6"
```
klemen1999: Based on my understanding we should be able to keep dynamic version import as we had so far. I don't believe keeping it hardcoded like this is required my `uv` or?

kozlov721: Yeah it does not need to be in `pyproject.toml`. I moved the version to a separate private `_version.py` module that is used by both `pyproject.toml` and `luxonis_train/__init__.py`. Doing it this way makes the package a bit more hardened because we're importing `pydantic_extra_types` in `luxonis_train/__init__.py` before defining `__version__`.

requirements.txt
```diff
@@ -1,24 +1,234 @@
...
+opentelemetry-proto==1.44.0
+opentelemetry-sdk==1.44.0
+opentelemetry-semantic-conventions==0.65b0
+optuna==4.9.0
+optuna-integration==4.9.0
+ordered-set==4.1.0
+packaging==26.2
+pandas==2.3.3
+pillow==12.3.0
+pillow-heif==0.21.0
```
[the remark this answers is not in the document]
kozlov721: Addressed [here](https://github.com/luxonis/luxonis-ml/pull/453)

requirements.txt
```diff
@@ -1,24 +1,234 @@
...
+filelock==3.32.0
+filetype==1.2.0
+flask==3.1.3
+flask-cors==6.0.5
+flatbuffers==25.12.19
+fonttools==4.63.0
+frozenlist==1.8.0
+fsspec==2024.6.1
+gcsfs==2024.6.1
+gdown==4.7.3
```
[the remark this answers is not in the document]
kozlov721: Addressed [here](https://github.com/luxonis/luxonis-ml/pull/453)

luxonis_train/_version.py
```diff
@@ -0,0 +1 @@
+__version__ = "0.4.6"
```
klemen1999: Out of curiosity: Is this more standard approach? Having a separate `_version.py` file? Or is there a specific reason why this was put out of the init?

pyproject.toml
```diff
@@ -26,19 +52,53 @@ luxonis_train = "luxonis_train.__main__:app.meta"
...
 [build-system]
 requires = ["setuptools", "wheel"]
 build-backend = "setuptools.build_meta"
 
+[tool.uv]
+# `opencv-python-headless` (via `albumentations`, `albucore`, `roboflow`) ships
+# the same top-level `cv2` as the `opencv-python` that `luxonis-ml` and
+# `grad-cam` require, so installing both leaves `cv2` a mix of the two decided
+# by install order. Drop the headless one; `opencv-python` satisfies everyone.
+override-dependencies = ["opencv-python-headless; sys_platform == 'never'"]
```
klemen1999: Currently `pip install -r requirements.txt` is failing because of:
```
ERROR: In --require-hashes mode, all requirements must have their versions pinned with ==. These do not:
    opencv-python-headless>=4.9.0.80 from https://files.pythonhosted.org/packages/9b/21/f6ef335f6e65724aa78b8d792b48d40a48c381715f1e62f5a5049e09d07e/opencv_python_headless-5.0.0.93-cp37-abi3-manylinux_2_28_x86_64.whl#sha256=ed709fdf9aa0bd1f2ed8549e71d19449b03a675bb581eb292285f6861953be37 (from albucore==0.0.24->-r requirements.txt (line 77))
```
This works for `uv` installs since we override dependencies in pyproject.toml but this doesn't have an effect on "pip install" path. Since we still offert and advertise "pip install" path (which I think we should keep) we need to fix and also keep testing this path since only manual test caught this now.

.github/workflows/ci.yaml
```diff
@@ -122,30 +134,25 @@ jobs:
...
         with:
-          python-version: "3.10"
-          cache: pip
+          version: "0.11.28"
+          enable-cache: true
 
-      - name: Install dependencies
-        run: pip install -e .[dev,aimet] --extra-index-url https://download.pytorch.org/whl/cu130
+      - name: Sync dependencies
+        run: uv sync --locked --extra aimet --group dev
```
klemen1999, line 144: On released PRs we are no longer installing latest released Luxonis-ml because we are using --locked here and the check below is checking if branch name is not release/.
Is this intended? Same happens or "tests" job

CONTRIBUTING.md
```diff
@@ -23,62 +12,91 @@ git clone git@github.com:luxonis/luxonis-train.git
...
 
-We use the [Epytext](https://epydoc.sourceforge.net/epytext.html) markup language for documentation.
-To verify that your documentation is formatted correctly, run the following command:
+We use the [Epytext](https://epydoc.sourceforge.net/epytext.html) markup
+language for documentation. To verify that your documentation is formatted
+correctly, run the following command:
 
 ```bash
-pydoctor --docformat=epytext luxonis_train
+uv run --only-group docs pydoctor --docformat=epytext luxonis_train
```
klemen1999: Let's use `--group docs` instead of `--only-group` so it just adds docs requirements to your existing env instead of uninstalling all the extra packages.

pyproject.toml
klemen1999: Most likely the fix should be that export_requirements.sh still generates requirements that keep installing both `opencv` and `opencv-headless`

luxonis_train/_version.py
kozlov721: There were some arguments for this (importing `__version__` without the entire package) but it's mostly just an AI-ism so I ended up reverting it. Version lives in `__init__` like before now.

klemen1999 (approved):
LGTM
For tracking: If one goes with `pip install -r requirements.txt` they should use **pip 23.3 or newer** since older have some bug in dependency resolver which results in failed pip installation.

PR #443, opened 2026-09-02
Pre-docs cleanup: fix the typo'd names and tighten the lint gate

kozlov721:
## Purpose

Prepare the repository for the Google docstyle migration and clean it
up in general. The PR mirrors the pre-docs cleanup in `modelconverter`
(luxonis/modelconverter#276) and the "Small breaking changes" PR in
`luxonis-ml` (luxonis/luxonis-ml#426). The docstyle migration follows
in a separate PR; it converts the docstrings that this PR makes
correct.

## Specification

- Add the `typos` pre-commit check with the same pin as the sibling
  repositories, and fix the 75+ typos it found.
- Rename the typo'd public names:
  - `reparametrize` → `reparameterize` (base class, module file,
    `GeneralReparameterizableBlock`, methods).
  - `AttentionRefinmentBlock` → `AttentionRefinementBlock`.
  - `mask_coeficients` → `mask_coefficients` (head output key, loss
    parameter, tests — renamed together, so the wiring holds).
  - PPLCNetV3 `detecion_blocks` → `detection_blocks`.
- Expand the ruff configuration to the modelconverter rule set:
  `ASYNC`, `EXE`, `G`, `PGH`, `SLOT`, full `PERF`, formatted docstring
  code examples, and a ban on `__future__` imports. Fix the new
  findings and bump the ruff hook to match the `~=0.16` dev pin.
- Fix all 49 pydoctor "documented parameter does not exist" warnings.
- Correct the stale paths and examples in the README.
- Exclude the stale `build` directory from pyright.
- Anchor the `head-branch` regexes in `.github/labeler.yaml`, so a
  branch such as `ci/release-automation` no longer gets the `release`
  label and no longer disappears from the release notes.
- Bump the `prettier` hook from v3.8.3 to v3.9.6.
- Format the Python code blocks in `README.md` and `examples/README.md`.
- Fix the "wit ha single key" typo in a node warning. The `typos` hook
  does not find it, because both words are correct.
- Stop the variant metaclass from calling `__init__` inside its
  `except NotImplementedError` handler. A node without variants often
  builds itself from a remote checkpoint. A network failure there
  chained onto the `NotImplementedError`, so the traceback opened with
  a variants error and hid the real cause.
- Add unit tests for the bounding-box helpers, the tuner parameter
  parsing, and the metadata decode pass-through.

Out of scope, with their own follow-ups:

- The remaining pydoctor warnings are format-level (`@license` fields,
  link targets). The docstyle migration rewrites those docstrings.
[...]

PR #444, opened 2026-09-02
Gate cognitive complexity with `complexipy`

kozlov721:
## Purpose

Add a cognitive-complexity gate. The gate enforces complexipy's
threshold of 15 on every function. The codebase passes the gate today.
The gate uses no baseline file and no per-function exemptions. This
pull request is stacked on #443.

## Specification

- Refactor the functions that exceed the limit. Each refactor extracts
  private helpers in the same module.
- The largest drops are:
  - `BaseNode.run` 71 -> 1
  - `LuxonisModel.__init__` 82 -> 13
  - `default_annotate` 65 -> 8
  - `FailOnNoTrainBatches.on_fit_start` 58 -> 1
  - `get_mlflow_logging_keys` 57 -> 0
- Type the extracted helpers with their real types instead of `Any`.
  `BaseAttachedModule.get_parameters` now declares
  `dict[str, Tensor | list[Tensor] | None]`. The `None` values for
  optional missing arguments were always possible.
- Run the `complexipy` pre-commit hook (v7.0.1) with
  `--suggest-refactors`. A failure then prints a refactor plan for each
  function that breaks the limit.
- Add `complexipy~=7.0` to the dev group. Add a "Cognitive Complexity"
  section to CONTRIBUTING.
- Export the metric reduction type as `DistReduceFx`. `MetricState` is
  public, but its `dist_reduce_fx` field carried a private type, so a
  user could pass a value but could not name its type.
- Add unit tests for these areas:
  - the no-train-batches message helpers
  - the config upgrade of a `.json` file
  - the loader collate function

- New dev dependency: `complexipy~=7.0`. There are no runtime
  dependency changes.
- Public names, parameters, and defaults stay identical. The diff adds
  one public name, `DistReduceFx`.
- The refactors fix nine latent bugs. Behavior differs from the base in
  these cases:
  - `BaseNode.run` feeds `forward` parameters named `y` and `z` from
    input packets 1 and 2. A dead store made the old code feed packet 0
    to all of `x`, `y`, and `z`. No node in the repository uses those
    names. A custom node that uses them now raises an `IndexError` if
    it receives fewer packets. The old code passed it packet 0.
  - `default_annotate` continues with the next image after it yields an
    empty record. The old code stopped the whole batch. The shipped
    path uses batches of one image. Its output does not change.
  - `instances_from_batch` shapes its empty placeholders like each
    payload tensor. The old code shaped them all like the bounding
    boxes.
  - `compute_iou_loss` with `mask_positive=None` and `target_scores`
    set now computes a loss. The old code crashed on a float mask.
  - `decode_text_metadata_labels` returns the raw array when any row
    holds an invalid code. It also decodes an all-zero padding row to
    an empty string. The old code stopped the decode when the first row
    held no characters, and then returned the raw array. The return
    value therefore changes dtype and shape for a padded batch.
  - `_get_train_dataset_name` reads the name from
    `LuxonisDataset.identifier`. The old code read a `dataset_name`
    attribute that does not exist, so it always returned `None`. The
    HubAI variant name changes from `<model>` to `<model>:<dataset>`.
    Only a `LuxonisLoaderTorch` supplies a name. A custom loader gives
    `None`.
  - `upgrade_config` parses a `.json` config file with `json.loads`.
    The old suffix test omitted the dot, so `yaml.safe_load` parsed
    every file. The `upgrade config` command uses the same loader now,
    and it writes JSON to a `.json` output.
  - The export uploads the modelconverter YAML file after it closes the
    file. The old code uploaded the file inside the `open` block,
    before the buffer reached the disk.
  - The tuner raises a `ValueError` for a malformed `*_subset` key that
    does not name augmentations. The old code raised a `KeyError`.
- Four messages change:
  - The wildcard-target error says "which label" instead of "which of
    the labels".
  - The "Adjusting parameters" log loses a double space.
  - The no-train-batches error loses a duplicate space.
  - The `resize_along` error interpolates the value. The old code
    printed the literal `{resize_along}`.

## Deployment Plan

None / not applicable - the gate activates through pre-commit and CI.

## Testing & Validation

- `uv run complexipy luxonis_train tests`: exit 0. Every function stays
  at or below 15.
- `uv run prek run --all-files`: all hooks pass, the new gate included.
- Unit suite: 569 passed. The 9 failures in `tests/unittests/test_losses`
  come from expired gcloud ADC credentials. They are environmental.
- pyright: 2 errors. The base branch gives the same 2 errors. Both come
  from the optional `aimet_torch` extra.
- `codecov/patch` reports 90.0% and fails against the default target.
  Project coverage rises from 94.63% to 95.11%.

## AI Usage

Assisted-by: Claude:claude-fable-5, Claude:claude-opus-5

Submitted code was reviewed by a human: NO

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01KXj7GZFCcdShtaTweRNH9j
https://claude.ai/code/session_012cAuJ2RUss7HZGyCvwLi4V
https://claude.ai/code/session_01C9WMoUvjBxLnqYrGJdNZrg

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected positional input handling for multi-input nodes.
  - Annotation generation now handles text-only outputs and images without predictions.
  - Mixed tensor and dictionary input batches now produce a clear validation error.
  - Improved EMA checkpoint restoration warnings and compatibility handling.
  - Improved configuration upgrades and duplicate-name handling.
  - Improved handling of segmentation labels without matching classification labels.

- **Documentation**
  - Added contributor guidance for cognitive-complexity checks and refactoring suggestions.

- **Developer Experience**
  - Added automated cognitive-complexity checks.
  - Expanded test coverage across core workflows and utilities.


klemen1999 (approved):
Generally LGTM, no real issues found by multiple agents

